PR #10, opened 2018-07-09
Implement Textarea

jhoerr:
Text areas and text input are essentially the same thing in Rivet. The bulk of the `Input` component was refactored to `common.tsx` and shared with `Textarea`.


src/components/Input/Textarea.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+            expect(cut.find('label').prop('htmlFor')).toEqual("the_id");
+        });
+        it('should apply custom class', () => {
+            const cut = shallow(<Textarea className="my-class" label="Label" />);
+            expect(cut.find('div.rvt-input').hasClass("my-class")).toEqual(true);
+        });
+    });
+    describe("Notes", () => {
+        it('should apply the note', () => {
+            const cut = shallow(<Textarea note="Note" label="Label" />);
```
jolamar: note attribute to me is not entirely obvious.. at first I thought it might be the default value of the textarea (is default value an attribute?) but then realized it is just a note about the text input. This is interesting and makes me wonder if we should add validation pieces to the components.. 1.) the rule for validation and 2.) the alert type and message for invalid fields..

src/components/Input/Textarea.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+            expect(cut.find('small').text()).toEqual("Note");
+        });
+        it('should associate the input with the note', () => {
+            const cut = shallow(<Textarea id="the_id" note="Note" label="Label" />);
+            expect(cut.find('textarea').prop('aria-describedby')).toEqual("the_id_note");
```
jolamar: yeah I think you're aiming for something like this https://rivet.uits.iu.edu/components/forms/text-input/#textarea-inline-validation-states maybe?

src/components/Input/Textarea.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+        });
+    });
+    describe("Inline Alerts", () => {
+        it('info style', () => { 
+            const cut = shallow(<Textarea info note="Note" label="Label" />);
+            expect(cut.find('textarea').hasClass("rvt-has-info")).toEqual(true);
+            expect(cut.find('.rvt-inline-alert').hasClass("rvt-inline-alert--has-info")).toBe(true);
```
jolamar: ah I spoke to soon! :)

src/components/Input/common.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <label htmlFor={inputId}>{props.label}</label>
+
+const noteFragment = (inputId: string, props: TextProps) =>
+    isInlineAlert(props)
+        ? inlineAlert(props)
+        : standardNote(inputId, props);
+
+type TextComponentGenerator = <T>(id:string, className: string, ariaDescribedBy: string, ariaInvalid:boolean, attrs: T) => JSX.Element; 
+
+export const renderInput = <T extends React.HTMLAttributes<HTMLElement>>(
```
jolamar: what does "T" stand for?

jhoerr: `T` is convention for a generic type.  It can be read as, "renderInput is generic for any type T that extends HTMLAttributes<...>".

PR #12, opened 2018-07-09
Implement radio button and basic form.

jhoerr:


jhoerr:
RadioButton and Checkbox are essentially the same thing, just with different `input[type="..."]`. We might look at refactoring/combining those components to guarantee similarity.

PR #14, opened 2018-07-10
Convert optional Rivet style attributes to string enumerations

jhoerr:
This PR is being opened early so that we can discuss changes and make course corrections as we go.

By way of guiding principles, optional styling properties should:
1. be optional in the props list
2. have names prefixed with `rvt` 
3. have a `"default"` type in addition to whatever style types.
4. have a comment/jsdoc that links to the relevant style documention in the Rivet docs. 

This allows optional styling properties to be excluded when not needed, easily discoverable, and explicitly set as "default" by developers (that is, the default behavior should not require their omission.) 

ex:
```
interface ButtonProps extends Rivet.Props {
    /** 
    * Optional Rivet style: a success/danger/plain button. 
    * See: https://rivet.uits.iu.edu/components/forms/buttons/#button-examples 
    */
    rvtStyle?: "success" | "danger" | "plain" | "default"
}
```


jhoerr:
Ok, this is ready to merge. @jolamar Would you mind taking a look and seeing if anything looks super problematic? There's a lot of little changes, apologies... :\

jwbennet:
Just curious, but why prefix the attributes with `rvt`?

jwbennet:
I had one other thought.  John wasn't in the meeting, but we had discussed a similar approach to prefixing attribute names with `rvt` a while back and decided against it.  That's the primary reason why I'm asking about it now.

jhoerr:
@jwbennet What approach did y'all end up taking? I'd like to learn from your experience.

jwbennet:
I should mention that the the discussion I am referencing in these posts were specifically about the rivet-react project and were already mostly reflected in the project as it was before the Typescript conversion began.

Some of the same arguments you make above were made during the last discussion about this and the arguments against it were broken down like this:

1. Compare peer projects which are doing similar things and you won't see many examples of prefixing.  There are some like `bsStyle` to avoid collisions with HTML attributes but most properties do not have prefixes.  Here are links to two popular Bootstrap projects which implement similar APIs and if you go through a number of components you will see the limited use of the pattern:
  * https://react-bootstrap.github.io/
  * https://reactstrap.github.io/ (this one does not use a prefix at all)
2. When there are collisions with HTML attributes, then it is best to document the interaction regardless.  If your component supports `rvtTitle` and `title` the end user would still need to reference documentation to understand how those are related.  For cases like this it seems better to not use the prefix since it was clearer to understand without it.
3. We also discussed that if you really need to support something like title but want to be explicit you could use something like an `html` prefix like React does in some examples:
  * https://reactjs.org/docs/dom-elements.html#htmlfor
4. I should mention that the original proposal was to prefix Rivet-React properties with `_` (e.g. `_style`) but that was quashed due to concerns that it would cause confusing that people were interacting with "private" properties due to the naming convention.

We did not discuss things like intellisense in our past discussion, but I'm not really seeing the value in prefixing for that use case either.  It appears that there are a number of optional parameters which may be noise in this case, but to me the first screenshot is actually easier to follow because I'd be able to start typing what I'm looking for and after 3-4 characters I would see if it exists.  You could imagine that a complex component (e.g. header) with a number of properties would be just as hard to navigate with all of the properties sharing the prefix.  I understand where you come from about not wanting people to have to read docs or demos, but there are cases where that is going to be required.  I'd imagine most components will not have any properties which would have serious conflicts that would need any documentation to explain how the properties work, but it is always best to have good documentation to explain the few properties which have more business logic associated with them.

jhoerr:
Thanks James, points well taken. Is the prefix the only sticking point on this PR?

jhoerr:
I know there's quite a lot to it, and don't want to rush you. I'm just looking to get a sense of whether y'all see other structural issues.

jwbennet:
I do prefer the changes made here to use enums with documentation instead of boolean properties.  Everything but the naming seems fine to me.

jhoerr:
Ok, thanks @jwbennet. @athill Would removing the prefix address your concerns, or do you have others?

jhoerr:
@jwbennet @jolamar If we remove prefix the properties, we could partially mitigate the discovery problem by only allowing a subset of the standard HTML button props. Perhaps just `disabled`, `active' (as part of the component props) and the `aria-*`attributes (provided by React typings.) What would y'all think of that?

athill:
I agree. Removing the prefixes addresses my major concern.

Thanks.

jhoerr:
Given that we don't want the prefix I'm thinking that we should remove support for ad-hoc HTML attributes outside of what is specifically defined in the component props. I'll refactor the Button as a proof of concept.

jhoerr:
The modified, no-prefix Button can be seen in fa8809c. Is this looking better?

jwbennet:
I'm not really sure why we would want to restrict the normal attributes users can pass down to the component.  It feels like it would be shortsighted to think we'll know what is needed in every use case. 
 I'm assuming the main change in the POC is that the button class no longer implements `React.ButtonHTMLAttributes<HTMLButtonElement>` so most of the normal attributes would not be passable when using TypeScript?

jhoerr:
Yep, that's correct, and I agree with you 100% that it'd be ideal to support normal HTML attributes.   

It seems that there's a natural tension between completeness (supporting normal HTML attributes) and discovery (making component attributes easy to identify). When I look at it through the lens of developers who (like me) are familiar-ish with Rivet and unfamiliar with React, discoverability feels most important. When I look at it through the lens of more seasoned developers, completeness feels most important.

I'm confident we can collectively figure out a way to thread the needle and I wonder if at this point it might be more easily worked out face-to-face with a whiteboard. Perhaps it's a good place to start on Monday? I feel like once we crack this nut and establish some conventions the rest will go pretty fast. 

I'm really glad we're talking about this stuff and I'm excited to be working with you all on it! I have a lot to learn!

jolamar:
Very sorry for not jumping on this discussion earlier! Been wrapped up in some other web projects this week. However, this is a very good PR and some good discussions around discoverability and customization. My main view is that Rivet React is usable to most developers. I see the benefit of being able to discover what a component can do by typing a `rvt` prefix and I think this kind of just-in-time education is valuable if it doesn't sacrifice something else important - I think in this case simplicity of just writing the attribute name out (so `title` instead of `rvtTitle`) and if `title` is already a native attribute then the component can support `htmlTitle` for the native title attribute. I think we should lean on people knowing some React standards like this and make sure the components can be customized fully (with all supported native attributes). We can't predict every way someone is going to use a component, but we should document the ways we intend it being used.  I think React Bootstrap docs are 💯 So I think some of the prefixes you have would work.. `rvtStyle`, `rvtSize` and maybe `rvtRole`. These all seem like enum attributes.. so maybe that's how we determine what gets prefixed? @jwbennet can you explain how Bootstrap decides to prefix and when they opt to prefix with `html` instead of `bs`?

jwbennet:
I was trying to find a GitHub issue explaining when react-bootstrap would prefix certain properties and found [this issue](https://github.com/react-bootstrap/react-bootstrap/issues/2829).  It looks like they are planning to remove the `bs` prefixes in their next major version.

jhoerr:
Test and Demo page are working for `Alert`. The tests were failing for the `DismissableAlert` because the `shallow`-wrapped `Alert` was not being rendered. One can use [`.dive()`](https://github.com/airbnb/enzyme/issues/947#issuecomment-302233681) to render the immediate child of a `shallow`-wrapped component and test assertions thereon.      ☄️ _The More You Know!_

src/components/Rivet.tsx
```diff
@@ -43,13 +43,17 @@ export interface Props {
...
 }
 
+export type Typescale = "base" | 12 | 14 | 16 | 18 | 20 | 23 | 26 | 29 | 32 | 36 | 41 | 46 | 52;
 export type Size = 'xxs' | 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'xxl';
 export const Sizes = [ 'xxs', 'xs', 'sm', 'md', 'lg', 'xl', 'xxl' ];
 export type Edge = 'top' | 'right' | 'bottom' | 'left';
 export const Edges = [ 'top', 'right', 'bottom', 'left' ];
+export type Display = "inline" | "inline-block" | "block" | "flex" | "flex-vertical-center";
+export type Border = "all" | Edge | Edge[]
+export type Hidden = "sm-down" | "md-down" | "lg-down" | "xl-down" | "xxl-down" | "sm-up" | "md-up" | "lg-up" | "xl-up" | "xxl-up"
```
jhoerr: 👍

src/components/Rivet.tsx
```diff
@@ -127,49 +131,86 @@ const parseRivetSpacing = (type, style?: Size | BoxStyle) =>
...
+});
+
+const legacyStandardDecorators = [
+    parseRivetMarginLegacy,
+    parseRivetPaddingLegacy,
+    parseRivetTypescaleLegacy,
+    parseRivetBorderLegacy,
+    parseRivetDisplayLegacy,
+    parseRivetHiddenLegacy
 ];
```
jhoerr: Is the plan to get rid of these *Legacy decorators when everything's migrated from `classify` to `rivetize`?

jwbennet: Yes, that is what I was thinking.

PR #20, opened 2018-07-19
Add components for the Rivet grid system

jwbennet:


src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { rivetize } from '../Rivet';
+
+const Grid : React.SFC<React.HTMLAttributes<HTMLDivElement>> = ({ children, className, ...attrs }) => (
+    <div className={classNames('rvt-grid', className)} {...attrs}>
+        {children}
+    </div>
+);
+Grid.displayName = 'Grid';
+
+export default rivetize(Grid);
```
burnumd: If you, like me, wondered why `<div class="rvt-grid">` wasn't just added to `Container` and why `Grid` needs its own component, it's (presumably) because you can nest `Grid` components.

src/components/Grid/Column.tsx
```diff
@@ -0,0 +1,60 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import { rivetize } from '../Rivet';
+
+type ColumnWidth = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+type AutoColumnWidth = ColumnWidth | boolean;
+
+interface ColProps {
+    last?: boolean;
+    lg?: AutoColumnWidth,
```
burnumd: There are a couple fields here that have comma line-endings instead of semicolons.

src/components/Grid/Grid.tsx
jwbennet: I added an alias named `Row` to the `Grid` component because I also found it a bit confusing, but I was attempting to align the components with the Rivet documentation.

burnumd: Thanks, that sounds like a helpful addition. I was curious, so I went over to the Rivet docs to see how the grid class was intended to be used in cases other than as the direct child of a div with the container class and saw that use case.

PR #28, opened 2018-07-23
Implement ButtonGroup component

jwbennet:
Adds the button group component described in https://rivet.uits.iu.edu/components/forms/buttons/#button-groups

### DIFF
--- a/src/components/Button/ButtonGroup.tsx
+++ b/src/components/Button/ButtonGroup.tsx
@@ -1,0 +1,25 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import * as Rivet from '../Rivet';
+
+interface ButtonGroupProps {
+  /**
+   * Right align the buttons in this button group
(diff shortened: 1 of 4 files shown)


src/components/Button/ButtonGroup.tsx
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * Right align the buttons in this button group
+   * @see https://rivet.uits.iu.edu/components/forms/buttons/#button-groups
+   */
+  right?: boolean;
+};
+
+const ButtonGroup: React.SFC<ButtonGroupProps & React.HTMLAttributes<HTMLDivElement>> = ({ className, children, right, ...attrs }) => {
+  const classes = classNames({
+    ['rvt-button-group']: true,
```
athill: I just thought of this, but are the square brackets required here?

jwbennet: No, they are not but I was including them explicitly so it would be easier to read since the keys would all be written the same way.

PR #42, opened 2018-07-24
 Implement Badge component

jwbennet:
Adds a Badge component and tests as described in https://rivet.uits.iu.edu/components/page-content/badges/

### DIFF
--- a/src/components/Badge/Badge.tsx
+++ b/src/components/Badge/Badge.tsx
@@ -1,0 +1,24 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import { rivetize } from '../Rivet';
+
+interface BadgeProps {
+    secondary?: boolean,
+    role?: 'default' | 'secondary',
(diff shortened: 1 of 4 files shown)


src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,24 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import { rivetize } from '../Rivet';
+
+interface BadgeProps {
+    secondary?: boolean,
```
burnumd: Is this `secondary` prop being used anywhere? I personally prefer this API versus having to specify when something should be the default.

jwbennet: I refactored the `secondary` property from a previous iteration to be called `role` instead and had missed removing this property.  I also prefer this style of property, but I went with the convention for now.  If we want to tweak that convention I would certainly agree with it.

PR #43, opened 2018-07-24
Add style guide and property documentation

jhoerr:
Create living component documentation and style guide using [React Styleguidist](https://react-styleguidist.js.org/) and [react-docgen-typescript](https://github.com/styleguidist/react-docgen-typescript). The style guide is split into sections to complement the Rivets component documentation.

This style guide subsumes and replaces the Demo app. On a subsequent branch we'll break the Demo app into component-specific examples and remove it. 

Component authors will need to add their component to the appropriate section in `styleguide.config.js` as well as create a `<Component>.md` file with usage examples. These example files have been added for `Alert` and `DismissibleAlert` to demonstrate how it's done.


package.json
```diff
@@ -38,6 +38,8 @@
     "@types/react-router-dom": "^4.2.7",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
+    "react-docgen-typescript": "^1.6.2",
+    "react-styleguidist": "^7.1.0",
```
jwbennet: I'd recommend removing the `^` at the front of these two dependencies to freeze it on specific versions.

jhoerr: Why these packages and not others?

PR #45, opened 2018-07-24
Component samples for style guide

jwbennet:
* Added component samples from the Rivet component documentation for each component we have converted to Typescript.
* Cleaned up several file names that caused issues with the style guide generation


jhoerr:
This rules!

src/components/Grid/Col.tsx
```diff
@@ -112,6 +112,10 @@ interface ColProps {
     xxl?: AutoColumnWidth;
 }
 
+/**
+ * This is a test
+ * @param props
+ */
 const Column : React.SFC<ColProps & React.HTMLAttributes<HTMLDivElement>> = ({ children, className, last, lg, md, pullSm, pullMd, pullLg, pullXl, pullXxl, pushSm, pushMd, pushLg, pushXl, pushXxl, sm, xl, xxl, ...attrs }) => {
```
athill: Do we want to rename the function for consistency?

jwbennet: Sure, I just made that change.

PR #54, opened 2018-07-26
#32 - Refactor InlineAlerts out of Input

### LINKED ISSUE #32: Implement InlineAlert (SFC)
* Refactor Input and TextArea to use InlineAlert

burnumd:
This change closes #32 
* Refactors `InlineAlert` out of the `Input` common utilities into its own component
* Documents and tests the inline alert functionality
* Implements `standalone` alerts for use outside of text inputs

Known Issues:
* `Alert` and `InlineAlert` use different values for `variant`. This conflict is a result of trying to preserve the API used in `Input` since Rivet itself uses two different variant schema to name alert classes vs. inline alert class names.


burnumd:
I am interested in opinions on where the Inline Alerts should go in the style guide. The Rivet docs split them between Alert and Text Input (even though the standalone inline alert appears under the Alert documentation, it uses the same variant naming scheme as the inline alerts that appear in the Input section). I have them in the Overlays section here, but I went back and forth on where they belong.

PR #57, opened 2018-08-08
Issues/55/modal

athill:


src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Button or buttons which will appear in the footer of the modal
+     */
+    controls: React.ReactElement<ButtonProps> | Array<React.ReactElement<ButtonProps>>;
+}
+
+
+class Modal extends React.PureComponent<ModalProps & React.HTMLAttributes<HTMLButtonElement>> {
+    
+    public static displayName = 'Modal';
```
jwbennet: Are you sure `displayName` is required?  I thought it was only needed for stateless components.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private shouldToggleDropdown = (event: ModalEvent) => {
+        if (event.isRightMouseClick() || event.isUnhandledKeyPress()) {
+            // If the user right clicks anywhere on the screen or they press an unhandled key do not close the menu
+            return false;
+        }
+
+        if (event.targets(ReactDOM.findDOMNode(this)) && !event.isKeyEvent()) {
+            // If the user clicks ortouches inside the modal do not close the menu
```
jwbennet: Typo here, should be `or touches`

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+
+    private closeModal = (event) => {
+        if (this.props.onDismiss) {
+            this.props.onDismiss();
+        }        
+    }
+
+    private shouldToggleDropdown = (event: ModalEvent) => {
```
jwbennet: This isn't a dropdown so this function should probably be renamed

src/components/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+    });    
+
+    it('should call the onDismiss function if the close button is clicked', () => {
+        const onDismiss = jest.fn();
+        const cut = mount(<Modal title={title} controls={okButton} onDismiss={onDismiss}>{body}</Modal>);
+        cut.find('button.rvt-modal__close').simulate('click');
+        expect(onDismiss.mock.calls.length).toBe(1);
+    });
+
+    // TODO: see if this can be fixed
```
jwbennet: This TODO should be addressed before merging this code

src/components/Modal/ModalEvent.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    public targets = (container: Element | Text | null) => {
+        return (container && container.contains(this.target) && container !== this.target);
+    }
+
+    public toString = () => {
+        return `ModalEvent { type: ${this.type}, which: ${this.which} }`;
+    }
+
+}
```
jwbennet: Missing EOF newline

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,142 @@
...
+     */
+    onDismiss?: Rivet.Action;
+    /**
+     * The content of the modal's header
+     */
+    title: React.ReactNode | string;
+    /**
+     * Button or buttons which will appear in the footer of the modal
+     */
+    controls: React.ReactElement<ButtonProps> | Array<React.ReactElement<ButtonProps>>;
```
jwbennet: I personally don't like passing arrays of React elements as properties.  Have you considered using components to denote what goes in the header/body/footer like other libraries use?

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,142 @@
...
+            this.handleEventRegistration();
+        }
+    }
+
+    public componentWillUnmount() {
+        this.eventHandler.deregister();
+    }    
+    
+    public render() {
+        const { children, className, controls, isOpen, onDismiss, title, id = Rivet.shortuid() } = this.props;
```
jwbennet: This destructuring should be in alphabetical order and you might consider also spreading the remaining properties to pass down.

src/components/Modal/Modal.tsx
jwbennet: On top of this it feels odd to be that we allow passing arbitrary controls then render and additional one if an `onDismiss` handler is provided.

athill: I'm not a fan of passing in arrays of elements either, but this was what was started before I began. I can change it.

As far as the "close" button, it is not in the control section at the bottom, it is in the upper right corner (confusing as its at the bottom of the component definition). Having it depend on a function being passed in is similar to what react-boostrap does, except they have an additional property (`closeButton`) and describe their `onHide` as "A callback fired when the header closeButton or non-static backdrop is clicked. Required if either are specified." Essentially, the current implementation says if there is a function to handle dismissing, you get a close button for free, as opposed to react-bootstrap, that has an additional prop that you can toggle *only if you pass in the function*, so both render the close button only if the close handler function is provided.

athill: Looking at separating out `ModalHeader` etc., we are currently setting an `aria-labelledby` in the parent that matches the `id` of the header. Would this best practice be delegated to the user or how would we enforce it?

burnumd: Yeah, I did the original control button array implementation before I left mostly as a way to quickly reuse an existing component and enforce the type, but I have no qualms about nested components for the controls. Also, what Andy said about the close button: it's confusing, but that's the way the Rivet documentation renders dismissible modals (as well as how react-bootstrap does it, as Andy pointed out).

athill: I'm thinking we could have the header baked into the Modal (to resolve the issue I mentioned above) but then use `ModalBody` and `ModalFooter` children to more closely match `react-bootstrap`. Thoughts?

athill: Also, I haven't pushed it yet, but I am moving the "close" button into the `header` tag to make it more apparent  where it will actually render

src/components/Modal/DropdownEvent.test.ts
```diff
@@ -0,0 +1,28 @@
+import { mount } from 'enzyme';
```
jwbennet: This file name is `DropdownEvent.test.ts`

src/components/Modal/Modal.md
```diff
@@ -0,0 +1,111 @@
...
+  
+  onCloseButtonClick(e) {
+    this.setState({
+      open: false
+    });
+  }
+}
+
+<DialogExample />
+```
```
jwbennet: I like the extended documentation here.  We should consider doing similar things for other components.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,129 @@
...
+class Modal extends React.PureComponent<ModalProps & React.HTMLAttributes<HTMLDivElement>> {
+
+    public static propTypes = {
+        children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node]),
+        className: PropTypes.string,
+        id: PropTypes.string,
+        isOpen: PropTypes.bool,
+        onDismiss: PropTypes.func,
+        title: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node, PropTypes.string]).isRequired
+    };
```
jwbennet: Just curious, but John mentioned the other day that he hoped this would be generated in the transpiled code without it being specified.  Did we happen to check into that at all?

src/components/Modal/ModalEvent.test.ts
```diff
@@ -0,0 +1,11 @@
...
+
+describe('ModalEvent', () => {
+    describe('handler', () => {
+        it('should return an instance with register and deregister methods', () => {
+            const callback = jest.fn();
+            const handler = ModalEvent.handler(callback);
+            
+        });
+    });
+});
```
jwbennet: EOF newline

src/components/Modal/ModalEvent.ts
```diff
@@ -0,0 +1,13 @@
...
+
+    public static handler = (callback) => {
+        return AbstractUserActionEvent.handler(callback, ModalEvent);
+    }
+
+    public isUnhandledKeyPress() : boolean {
+        return this.isKeyEvent() && !this.isEscapeKeyPress();
+    }
+
+}
```
jwbennet: EOF newline

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+    protected handleEventRegistration() {
+        if(this.props.onDismiss && this.props.isOpen) {
+            this.eventHandler.register();
+        } else {
+            this.eventHandler.deregister();
+        }
+    }    
+
+}
```
jwbennet: I'm wondering if we could expose the `ModalBody` and `ModalControls` as `Modal.Body` and `Modal.Controls` like we're doing with the header?

src/components/Modal/Modal.tsx
jhoerr: I'm not sure this is something we'll get for free from TS. Investigating our options now. Hopefully there's a library or something that can autogenerate propTypes given some interface.

src/components/Modal/DropdownEvent.test.ts
athill: Fixing. Thanks.

src/components/Modal/ModalEvent.test.ts
athill: Fixing. Thanks.

src/components/Modal/ModalEvent.ts
athill: Fixing. Thanks

src/components/Modal/Modal.tsx
athill: I'll give it a shot. I hadn't noticed we'd found a way to do that.

src/components/Modal/Modal.test.jsx
```diff
@@ -0,0 +1,247 @@
...
+    const defaultTitle = 'modal title';
+    const okButton = <Button key="ok" id="ok-button">OK</Button>;
+    const defaultChildren = (
+        <div>
+            <Modal.Body>Lorem ipsum dolor sit amet, consectetur adipiscing elit.</Modal.Body>
+            <Modal.Controls>{okButton}</Modal.Controls>
+        </div>   
+    );
+
+    const component = (overrides={}) => {
```
jwbennet: I missed this before, but this helper function makes these tests significantly different from all of the others and while it does make this test have a little bit less code it makes them a lot less readable.  I'll approve the change as-is but this might be something we consider refactoring before release.

PR #58, opened 2018-08-10
Convert Dropdown component to Typescript

jwbennet:
Converts the Dropdown component to TypeScript and adds tests.

Note: The original Dropdown component was not removed since it is still used in the older Header component and it seemed like it would be easiest to remove it when we remove the Header.


jhoerr:
🙌

jhoerr:
The props need some documentation -- I'll add it.

src/components/Dropdown/DropdownGroup.tsx
```diff
@@ -1,6 +1,10 @@
 import * as React from 'react';
 
 interface DropdownGroupProps {
+    /**
+     * Optional `header for the item group.
```
athill: Unmatched back tick

PR #62, opened 2018-08-13
Extract BaseEvent class

athill:
* Extract most of DropdownEvent into BaseEvent
* Move `keys` into BaseEvent

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -36,7 +36,7 @@ export class Dropdown extends React.PureComponent<DropdownProps & React.HTMLAttr
         this.toggleButton = React.createRef();
         this.handleClickOutside = this.handleClickOutside.bind(this);
         this.toggleDropdown = this.toggleDropdown.bind(this);
-        this.eventHandler = DropdownEvent.handler(this.handleClickOutside, DropdownEvent);
+        this.eventHandler = DropdownEvent.handler(this.handleClickOutside);
     }
 
     public componentDidMount() {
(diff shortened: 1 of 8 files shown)


jhoerr:
Primo refactor, @athill

src/components/util/BaseEvent.ts
```diff
@@ -0,0 +1,61 @@
+
+
+export default abstract class BaseEvent {
```
jwbennet: How about renaming this to something more descriptive?  Maybe `AbstractUserActionEvent`?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -36,7 +36,7 @@ export class Dropdown extends React.PureComponent<DropdownProps & React.HTMLAttr
         this.toggleButton = React.createRef();
         this.handleClickOutside = this.handleClickOutside.bind(this);
         this.toggleDropdown = this.toggleDropdown.bind(this);
-        this.eventHandler = DropdownEvent.handler(this.handleClickOutside);
+        this.eventHandler = DropdownEvent.handler(this.handleClickOutside, DropdownEvent);
```
jwbennet: How about we provide another version of this function in `DropdownEvent` which provides the second argument so the code that uses it does not have to?

athill:
@jwbennet Requested changes made

src/components/util/AbstractUserActionEvent.test.ts
```diff
@@ -0,0 +1,67 @@
...
+const createMouseEvent = (buttonCode) => ({
+    type: 'click',
+    which: buttonCode
+}) as MouseEvent;
+
+const createTouchEvent = () => ({
+    type: 'touchstart'
+}) as TouchEvent;
+
+describe('DropdownEvent', () => {
```
jwbennet: This probably needs to be changed to `AbstractUserActionEvent`

PR #65, opened 2018-08-14
Convert Table component to Typescript

jwbennet:
Converts the Table component to TypeScript and adds tests.

This removes the TableRow component and does not replace it.  It did not seem especially useful in a general case.


src/components/Table/Table.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+        });
+        it('should pass attributes through', () => {
+            const cut = mount(<Table id="the_id" />);
+            expect(cut.find('table').prop('id')).toEqual('the_id');
+        });
+        it('should apply custom classes', () => {
+            const cut = mount(<Table className="foo" />);
+            expect(cut.find('table').hasClass('foo')).toBe(true);
+        });
+    });
```
athill: Shouldn't you be testing the classes applied or not based on the variants?

PR #74, opened 2018-08-20
/issues/72/select

athill:
Implement the `Select` component

### DIFF
--- a/src/components/Input/Select.md
+++ b/src/components/Input/Select.md
@@ -45,7 +45,7 @@
        label="Type"
        note={<> You must pick a <strong>Type</strong> of meat. </>}
        margin={{bottom: 'md'}}>
-    <option value="Steak">Choose an option...</option>
+    <option value="">Choose an option...</option>
     <option value="Steak">Steak</option>
     <option value="Chops">Chops</option>
     <option value="Ribs">Ribs</option>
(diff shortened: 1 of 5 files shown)


src/components/Input/Select.md
```diff
@@ -0,0 +1,67 @@
+### Default Style
+
+```jsx
+<Select type="text" name="text-demo" label="Select Input" note="This is a note." margin={{ bottom: 'md' }}>
+    <option>Choose an option...</option>
+    <option value="Option One">Option One</option>
+    <option value="Option One">Option Two</option>
```
burnumd: The values here should probably be updated to match their text.

src/components/Input/Select.md
```diff
@@ -0,0 +1,67 @@
...
+### Select Inline Validation
+
+```jsx
+<Select type="text"
+       name="select-valid-state"
+       variant="valid"
+       label="Type"
+       note={<> You must pick a <strong>Type</strong> of meats. </>}
+       margin={{bottom: 'md'}}>
+    <option value="Steak">Choose an option...</option>
```
burnumd: Should this option not have a value (same with the other validation examples)?

src/components/Input/Select.md
athill: I copied  the options directly from the Rivet site. I started to do them significantly differently (i.e., with meaningful examples), but wasn't sure how much we wanted to match their examples, so I thought I'd start there.

burnumd: Sounds like it's time to file a bug against Rivet. 😜

src/components/Input/Select.md
athill: Actually, I would think it should have a value of empty string, as by default the value is the child of the option (i.e., all the `value` attributes that are the same as the child are redundant). Otherwise, any validation would have to check for the string "Choose an option..."

PR #75, opened 2018-08-20
Mobile Header

jhoerr:
Added a `HeaderDrawer` component and event handling for rendering identity and navigation elements for small screen (mobile) displays.

We added a `HeaderCollapse` which essentially behaves like any other collapse and would be suitable for use in an Accordion if we ever want to implement such a thing.

Many of the tests are for verifying `HeaderDrawer` visibility and behavior. At this point the Dropdown, Modal, and HeaderDrawer are all using more or less the same boilerplate code for handling click and keyboard events. We should look at refactoring this boilerplate code, or finding some library that provides this logic and doing away with the code (and accompanying tests) altogether. Deleted code is debugged code!


athill:
This looks good overall, but there are several issues with the stylesheet. `HeaderCollapse` and other things have no documentation (for, e.g. `ModalBody` I provided a brief description and linked to the primary component). 
Also, not sure now I feel about the mobile functionality not working in the stylesheet. I would be completely against it not working, but I admit that the examples on the Rivet site don't work either.

Can we provide a fully functional example?

athill:
Also, we could bring over some of the verbiage from the Rivet site to the style guide.

jhoerr:
@athill Gotcha. The mobile example works but, due to Rivet styles, the drawer appears all the way up at the top of the page.
@jwbennet Perhaps we can include in the style guide the styling fix for relative drawer placement?

athill:
@jhoerr Ah yes, I see it when I scroll up. Thanks for pointing that out.

jhoerr:
In re: style guide documentation, I think we might also limit the `Header` component docs to `Header`, `HeaderIdentity`, `HeaderNavigation`, and `HeaderMenu`. The rest are implementation details not consumed by developers.

jhoerr:
@jwbennet What would you think of applying the drawer positioning fix to the style guide docs? If you can tackle that I’ll work on the other documentation Andy requested.

jwbennet:
Sure, I just added some customized CSS which makes the header drawer render a bit better in the style guide.

jhoerr:
@athill I believe the requested changes have been applied. Have a look and us know if it needs others.

PR #83, opened 2018-08-22
Remove dot notation from Header and Modal

### LINKED ISSUE #81: Inconsistent use of dot notation in component names
When are using dot notation inconsistently when naming components. I'd like to see if we can come to some agreement on a unified approach. I'm tagging this as for Milestone 0.3.0 because it will be an (avoidable) breaking change if we wait.

**Problem Statement**

There are several complex components, including Header and Modal, in which we use dot notation to express a hierarchical relationship. There are two primary benefits to this approach: 

1. it simplifies `import` statements;
2. it visually implies that e.g. `Modal.Body` necessarily 'goes with'  a `Modal` and should only be used within a `Modal`.  That is, a `Modal.Body` is not intended to be used as a standalone component.

```jsx
import Modal from './Modal'

<Modal>
  <Modal.Body>...</Modal.Body>
  <Modal.Controls>...</Modal.Controls>
</Modal>

import Header from './Header'

<Header>
  <Header.Identity />
  <Header.Navigation>
    <Header.Menu>...</Header.Menu>
  </Header.Navigation>
</Header>
```

There are multiple components that do not use dot notation but are are similarly structured in that the 'child' components should not be used without a corresponding 'parent' component. Examples include `Dropdown`/`DropdownGroup`, and `Container`/`Grid`/`Col`:

```jsx
<Dropdown>
  <DropdownGroup>...</DropdownGroup>
</Dropdown>

<Container>
  <Grid>
    <Col>...</Col>
  </Grid>
</Container>
```

Another example is `Button`/`ButtonGroup`/`SegmentedButton`, though the structure is slightly different. In this case, the 'parent' component should not be used absent a corresponding 'child' component:

```jsx
<ButtonGroup>
  <Button />
  <Button />
</ButtonGroup>

<SegmentedButton>
  <Button />
  <Button />
</SegmentedButton>
```

**Proposed Solution 1: Strictly use dot notation everywhere.**

Pros: simplified imports across the board, and a clear indication of the 'primary' component in the hierarchy.
Cons: `rivetize` may not be able to maintain the statically-applied dotted children (see [Header](https://github.com/indiana-university/rivet-react/blob/develop/src/components/Header/Header.tsx#L41-L43) for example of this.

Suggested changes:

```jsx
<Dropdown>
  <Dropdown.Group>...</Dropdown.Group>
</Dropdown>

<Button.Group>
  <Button />
  <Button />
</Button.Group>.

<Container>
  <Container.Grid>
    <Container.Col>...</Container.Col>
  </Container.Grid>
</Container>
```

**Proposed Solution 2: Remove dot notation from Header and Modal**

Pros: `rivetize` continues to work without issue.
Cons: More verbose imports.

Suggested changes:

```jsx
import Modal, ModalBody, ModalControls from './Modal'

<Modal>
  <ModalBody>...</ModalBody>
  <ModalControls>...</ModalControls>
</Modal>

import Header, HeaderIdentity, HeaderNavigation, HeaderMenu from './Header'

<Header>
  <HeaderIdentity />
  <HeaderNavigation>
    <HeaderMenu>...</HeaderMenu>
  </HeaderNavigation>
</Header>
```

jhoerr:
Resolves #81

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -3,7 +3,6 @@ import * as React from 'react';
 import { findFirstChildOfType } from '../util/childUtils';
 import Drawer from './HeaderDrawer';
 import Identity from './HeaderIdentity';
-import Menu from './HeaderMenu';
 import Navigation from './HeaderNavigation';
 
 interface HeaderProps {
@@ -14,15 +13,9 @@ interface HeaderProps {
     title: string;
 }
 
-interface Header extends React.SFC<HeaderProps & React.HTMLAttributes<HTMLDivElement>> {
-    Identity?: typeof Identity;
-    Menu?: typeof Menu;
-    Navigation?: typeof Navigation;
-}
-
 const componentClass = "rvt-header";
 
-const HeaderComponent: Header = ({ children, className, title, ...attrs }) => {
+const HeaderComponent: React.SFC<HeaderProps & React.HTMLAttributes<HTMLDivElement>> = ({ children, className, title, ...attrs }) => {
     const identity = findFirstChildOfType(children, Identity.displayName);
     const navigation = findFirstChildOfType(children, Navigation.displayName);
     return (
@@ -56,8 +49,5 @@ const HeaderComponent: Header = ({ children, className, title, ...attrs }) => {
 }
 
 HeaderComponent.displayName = 'Header';
-HeaderComponent.Identity = Identity;
-HeaderComponent.Menu = Menu;
-HeaderComponent.Navigation = Navigation;
 
 export default HeaderComponent;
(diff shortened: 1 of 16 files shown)


jhoerr:
Resolve #81

src/components/Header/Header.tsx
```diff
@@ -3,7 +3,6 @@ import * as React from 'react';
 import { findFirstChildOfType } from '../util/childUtils';
 import Drawer from './HeaderDrawer';
 import Identity from './HeaderIdentity';
-import Menu from './HeaderMenu';
 import Navigation from './HeaderNavigation';
```
burnumd: Can we also remove this import for `Navigation`?

Issue #78, opened 2018-08-21
Remove the "default" option from various variant properties


jwbennet:
There are a number of components which use the string "default" to signify they should use default styling, but it would probably be better to remove those properties and just check for `undefined` instead.

jwbennet:
Fixed by #87

PR #84, opened 2018-08-23
Add some propTypes and inline some interfaces

athill:


src/components/Alert/common.ts
```diff
@@ -17,9 +19,20 @@ export interface StatefulAlertProps {
     onDismiss?: Rivet.Action
 }
 
+export const statefulPropTypes = {
+    variant: PropTypes.oneOf(['info', 'error', 'message', 'success']).isRequired,
```
jwbennet: Since you took the time to alphabetize the enum type we might as well have it alphabetized here as well.

src/components/Badge/Badge.tsx
```diff
@@ -15,6 +16,12 @@ interface BadgeProps {
     variant?: 'action' | 'error' | 'success' | 'warning';
 }
 
+
+const propTypes = {
+    role: PropTypes.oneOf(['default', 'secondary']),
+    variant: PropTypes.oneOf(['', 'action', 'error', 'success', 'warning'])
```
jwbennet: The `default` and empty string options should be removed now.

src/components/Badge/Badge.tsx
```diff
@@ -15,6 +16,12 @@ interface BadgeProps {
...
+};
```
jhoerr: The `'default'` and `''` options were removed in #78. Can you remove them here?

src/components/Button/Button.tsx
```diff
@@ -62,6 +72,10 @@ export const Button: React.SFC<ButtonProps & React.ButtonHTMLAttributes<HTMLButt
     </button>
   );
 };
+
```
burnumd: These statements don't need whitespace around them (and similar whitespace isn't applied elsewhere)

athill:
Requested changes have all been addressed. Thanks for the feedback.

PR #90, opened 2018-08-23
Improve Button docs and examples

jhoerr:


src/components/Button/Button.tsx
```diff
@@ -54,24 +53,28 @@ const buttonModifierAndStyle = (variant?: VariantType, modifier?: ModifierType)
...
-const buttonSize = (size?: SizeType) => (size !== undefined ? `${buttonClass}--${size}` : None);
+const buttonSize = (props: ButtonProps) => 
+  props.size
+  ? `${buttonClass}--${props.size}` 
+  : '';
 
 export const Button: React.SFC<ButtonProps & React.ButtonHTMLAttributes<HTMLButtonElement>> = 
-  ({ className, children, id = Rivet.shortuid(), innerRef, modifier, onClick, size, variant, ...attrs}) => {
-  const classes = classNames(buttonModifierAndStyle(variant, modifier), buttonSize(size), className);
+  ({ className, children, id = Rivet.shortuid(), innerRef, onClick, ...props}) => {
```
athill: Naming things. I agree `...props` makes more sense than `...attrs`, as that's what they are called in React. My hesitation is that we sometimes use `props` for all the props passed to a component (not destructured). I actually generally use `...others` in my personal projects, but I'm not claiming that's a good name. I'm fine with `...props` over `...attrs`, but thought I'd point out the overload.

jwbennet: This should be changed back to `attrs` since that's the standard the rest of the project uses

jwbennet: I'll also add that I really prefer spreading the props that this component actually uses.  I understand the workaround you're doing here, but if I was looking at this component to figure out what it did the props line is one of the first places I look.

src/components/Button/SegmentedButton.md
```diff
@@ -1,20 +1,33 @@
...
+    <Button onClick={() => { console.log('Click 3!') }}>Primary Three</Button>
+</SegmentedButton>
+```
+
+### Secondary Modifier
+```jsx
+<SegmentedButton>
+    <Button modifier="secondary" onClick={() => { console.log('Click 1!') }}>Secondary One</Button>
+    <Button modifier="secondary" onClick={() => { console.log('Click 2!') }}>Secondary Two</Button>
+    <Button modifier="secondary" onClick={() => { console.log('Click 3!') }}>Secondary Three</Button>
```
jwbennet: This example is fine, but nearly every other example in the project shows how to emulate the examples from the Rivet core documentation so this one does not match up with those.

src/components/Button/Button.tsx
jhoerr: Thanks for the feedback -- I'll take another swing.

PR #94, opened 2018-08-24
Extract all <svg> elements to new util/RivetIcons. Update usage.

jhoerr:


athill:
I like the idea of extracting these out. I just wonder if it might work to extract them to components. I always pause before using functions in component rendering that essentially return components.

jhoerr:
@athill I made them all functions for consistency because sometimes some of them need `classNames`. I can change them to `SFC`s just the same.

athill:
@jhoerr Just a thought. Curious if others have opinions.

jhoerr:
Would it make sense to leave the components in a single `src/util/RivetIcons.tsx` file, or have a new Icons namespace, e.g. `src/Icons/{Info, Close, Warning,...}.tsx`?

athill:
I think it would be overkill to have a separate file for each icon.

jhoerr:
I'll name the `SFC`s as `SvgInfo`, `SvgClose`, `SvgWarning`, etc. Cool?

src/components/util/RivetIcons.tsx
```diff
@@ -0,0 +1,88 @@
+import * as React from 'react'
+
+export const svg16 = (className: string = "", node: React.ReactNode) =>
+    <svg role="img" xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" className={className}>
+        {node}
+    </svg>
+
+/****************
```
burnumd: Instead of these block comments, would you consider using

```javascript
//#region My region name
... some code
//#endregion
```

Since most code editors (including VSCode and IntelliJ) recognize that and then support code folding, which we wouldn't get with block comments.

src/components/Alert/Alert.tsx
```diff
@@ -2,6 +2,7 @@ import * as classNames from 'classnames';
 import * as React from 'react';
 
 import * as Rivet from '../util/Rivet';
+import * as svg from '../util/RivetIcons';
```
burnumd: Is there a reason to `import *` instead of just importing the icon you need? All of the imports seem to be this way, so I'm not sure if there's a typescript-y reason for it.

src/components/util/RivetIcons.tsx
```diff
@@ -0,0 +1,88 @@
+import * as React from 'react'
+
+export const svg16 = (className: string = "", node: React.ReactNode) =>
```
burnumd: This name is kind of confusing (you have to look into the code of the rendered component to see that the `16` in the name refers to the size). Can it just be named `svgBase` or something similar?

src/components/Alert/Alert.tsx
jhoerr: I'm changing them all to SFCs which will be imported singly. Stay tuned.

src/components/util/RivetIcons.tsx
jhoerr: Will change, thanks for the suggestion.

burnumd:
I think I'd prefer them to be SFCs that you can render with, e.g., `<MenuIcon className="my-additional-class">` I definitely don't like the idea of including `Svg` in the component name because it exposes the implementation.

Personally, my ideal scenario is I render `<Icon name="success" className="my-additional-class" />`, and I think that might protect us better once Rivet extracts icons to a separate library, as well.

src/components/util/RivetIcons.tsx
burnumd: We may also want to consider making this size customizable (maybe not in this PR since there isn't an outstanding use case for it), but it's another reason not to include the size in the constant name.

jhoerr: Changing to SFC with DefaultProps, which should make the sizing adjustable via props.

jhoerr:
@burnumd I've adopted your suggested structure -- can you take another look? The tridents are not customizable because they're purpose-specific, but the rest will apply standard SVG properties.

jwbennet:
I saw John had a question earlier about Rivet icons being in a separate library.  Here it is: https://github.iu.edu/UITS/rivet-icons

burnumd:
@jhoerr 
> > I think that might protect us better once Rivet extracts icons to a separate library
> 
> Can you elaborate?

Sure, my thinking is that if we also wanted to break out icons to a separate library when Rivet does so, only some imports would change, and the API won't change even if Rivet's does. For example, if, in the split, Rivet renames `menu` to `hamburger` for whatever reason, the switch on our side can render the menu icon in either the `menu` or `hamburger` case and our end users won't have to change any of their code. Same for if Rivet introduces additional styling attributes on icons or introduces font glyphs or something, we can just add those without forcing our end users to adopt a new schema for icons and shields them from the underlying implementation.

PR #100, opened 2018-08-24
Don't publish gh-pages from develop

jhoerr:
Now that we've merged 0.3.0 (yay!), `master` should be the exclusive source for `gh-pages`.


jwbennet:
💯

PR #101, opened 2018-08-24
Refactor Footer and improve docs 

jhoerr:
Removed dependencies on `Rivet.Nav` and `react-router-dom`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "rivet-react",
-  "version": "0.3.0",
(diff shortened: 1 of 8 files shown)


jhoerr:
Removing the `React.Nav` class and `Footer.nav` prop does introduce an API change.

jhoerr:
I think that'd be appropriate.  Would it make sense to release this now as 0.4.0 and publishing the remaining doc updates as 0.4.x?

PR #105, opened 2018-08-28
Improve Input docs; refactor 

jhoerr:
I did a pretty big refactor to tidy up src/Input/common.tsx, remove some of the type verbosity, and (I hope) make things a little more React-y.

### DIFF
--- a/src/components/Input/Input.md
+++ b/src/components/Input/Input.md
@@ -1,11 +1,19 @@
-### Default Style
+Text inputs are the basic building blocks forms. They allow users to enter various types of data into web-based forms.
+
+View the [Rivet documentation for Text Inputs](https://rivet.uits.iu.edu/components/forms/text-input/).
+
+### Text input example
+
+Use the `note` property to provide contextual information to the user.
 
 ```jsx
 <Input type="text" name="text-demo" label="Text Input" note="This is a note." margin={{ bottom: 'md' }} />
 <Input type="text" name="text-demo-disabled" label="Text Input (Disabled)" disabled />
 ```
 
-### Input Inline Validation
+### Inline Validation
+
+Use the `variant` property along with a `note` to provide validation feedback to the user.
 
 ```jsx
 <Input type="text"
--- a/src/components/Input/Select.md
+++ b/src/components/Input/Select.md
@@ -1,4 +1,10 @@
-### Default Style
+The select element creates a dropdown that allows users to choose one item from a list.
+
+View the [Rivet documentation for Select](https://rivet.uits.iu.edu/components/forms/select-input/).
+
+### Select Example
+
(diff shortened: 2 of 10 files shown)


src/components/Input/Input.md
```diff
@@ -1,11 +1,19 @@
-### Default Style
+Text inputs are the basic building blocks forms. They allow users to enter various types of data into web-based forms.
```
athill: Not sure what you're trying to say here, but "Text inputs are the basic building blocks forms" does not make sense to me.

src/components/Input/Select.md
```diff
@@ -1,4 +1,10 @@
-### Default Style
+The select element creates a dropdown that allows users to choose one item from a list.
```
athill: Not sure if anything prevents using this for a multi-select. I didn't think to try.

src/components/Input/common.tsx
```diff
@@ -4,47 +4,50 @@ import * as React from 'react';
...
+        }
         return (
-            <div className={classNames('rvt-input', props.className)}>
-                {labelFragment(inputId, props)}
-                {inputGenerator <T> (inputId, inputClass, ariaDescribedBy, ariaInvalid, props)}
-                {noteFragment(noteId, props, note)}
+            <div className={classNames('rvt-input', className)}>
+                {labelFragment(id, label)}
+                {inputGenerator (inputProps)}
+                {noteFragment(noteId, variant, note)}
```
athill: Not clear why these aren't components.

src/components/Input/Input.md
jhoerr: These intro statement are all verbatim from the Rivet docs.

src/components/Input/common.tsx
jhoerr: They could be. I didn't think about that. Separate branch, maybe?

src/components/Input/Input.md
athill: Well, there you go.

src/components/Input/common.tsx
athill: I would prefer it, but not a blocker. Just didn't know if you had reasoning behind it.

jhoerr: Lemme see if I can knock that out real quick

jhoerr: The `labelFragment` just renders a `<label>`, so that was easy. The noteFragment() now just renders one of an `<InputAlert>` or a simple `<small>` block.

src/components/Input/Input.md
burnumd: This seems like a case where we should deviate from the Rivet docs and write something that makes sense and file an issue with Rivet to fix the documentation there and then rectify it if/when it gets fixed.

PR #110, opened 2018-08-28
Remove Section component

jhoerr:
This `Section` component is simply a `rivetize`d version of the <section> element. This component is a historical artifact and we do not currently do this for other simple elements (e.g. `a`, `p`, `span`, `div`).

This PR removes the Section component. We should consider implementing and documenting a general method for rivetizing simple elements, perhaps for milestone 0.5.0, or 0.4.0 as a stretch goal.


jwbennet:
I won't block this PR but I feel it is premature at this point.

jhoerr:
@jwbennet Can you say more about what feels premature about it? It will be harder to remove it once it's in use.

jwbennet:
If this is something people find useful I'm not sure why we'd want to remove it.  We have not had the chance to build something real with it yet so assuming it will not be useful to people feels premature at this point.  I do think we need to be careful about which components we include, but this one was brought in from a version of an existing library so at this point I'm assuming it has found use.

jhoerr:
I agree that it's useful. I think the problem is it's a one-off convenience, and we don't have corresponding components for all the other elements that would also be useful (and arguably moreso): `Div`, `Span`, `A`, `P`, `Img`, etc. I'm game for a converastion about creating components for all those. But, if we know for sure that we don't, then I don't see why we would keep `Section`.

PR #126, opened 2018-09-13
Add breadcrumb support

jwbennet:
https://rivet.uits.iu.edu/components/navigation/breadcrumb/

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.md
+++ b/src/components/Breadcrumbs/Breadcrumbs.md
@@ -12,9 +12,9 @@ View the [Rivet documentation for Breadcrumbs](https://rivet.uits.iu.edu/compone
 </Breadcrumbs>
 ```
 
-### Call out example
+### Callout example
 
-There is also a call out modifier class that adds a small amount of padding and a light gray background. This is useful for when you need to draw more attention to the breadcrumb.
+There is also a call-out modifier class that adds a small amount of padding and a light gray background. This is useful for when you need to draw more attention to the breadcrumb.
 
 ```jsx
 <Breadcrumbs variant="call-out">
--- a/src/components/Breadcrumbs/Breadcrumbs.md
+++ b/src/components/Breadcrumbs/Breadcrumbs.md
@@ -1,0 +1,25 @@
+A navigation component that indicates the user's current location in the nav hierarchy.
+
+View the [Rivet documentation for Breadcrumbs](https://rivet.uits.iu.edu/components/navigation/breadcrumb/).
+
+### Default breadcrumb
+
+```jsx
+<Breadcrumbs>
+    <a href="#">Home</a>
+    <a href="#">Files</a>
+    my-file.txt
+</Breadcrumbs>
+```
+
+### Call out example
+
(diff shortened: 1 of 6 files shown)


src/components/Breadcrumbs/Breadcrumbs.md
```diff
@@ -0,0 +1,25 @@
...
+
+```jsx
+<Breadcrumbs>
+    <a href="#">Home</a>
+    <a href="#">Files</a>
+    my-file.txt
+</Breadcrumbs>
+```
+
+### Call out example
```
jolamar: Callout*

src/components/Breadcrumbs/Breadcrumbs.md
```diff
@@ -0,0 +1,25 @@
...
+
+There is also a call out modifier class that adds a small amount of padding and a light gray background. This is useful for when you need to draw more attention to the breadcrumb.
```
jolamar: change 
>There is also a call out modifier class

to

>There is a `call-out` variant

src/components/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+              </Breadcrumbs>);
+            expect(cut.find('nav > ol.rvt-breadcrumbs > li').length).toEqual(2);
+        });
+        it('should mark the last child as the current page', () => {
+            const cut = mount(
+              <Breadcrumbs>
+                <a href="#">Link One</a>
+                Link Two
+              </Breadcrumbs>);
+            expect(cut.find('nav > ol.rvt-breadcrumbs > li').last().prop('aria-label')).toEqual('Current page');
```
jolamar: I believe the current page element should also have `aria-current="page"`

jolamar: I actually think `aria-label="Current page"` is fine

jolamar: https://www.w3.org/TR/wai-aria-practices-1.1/examples/breadcrumb/index.html here is where it mentions aria-current though

src/components/Breadcrumbs/Breadcrumbs.md
burnumd: I don't know that the call out -> callout changes are blocking. The Rivet docs themselves are inconsistent in how they refer to it. It's "Callout" in the header for that section, but "call out" and "call-out" (implying it's two words) in the actual example and code. I think we should hew towards the stylizing in the code and file an issue in Rivet to make the documentation consistent.

jolamar: I agree we should be consistent and that this is not a blocking issue. FWIW foundation uses `.callout` and `Callout` https://foundation.zurb.com/sites/docs/callout.html but I really don't think this will offend anyone if we name it `Call out`

src/components/Breadcrumbs/Breadcrumbs.test.tsx
jwbennet: Thanks for pointing that out.  I was working from an old breadcrumbs example which used `aria-label` but I changed it to `aria-current`.

jwbennet:
@jolamar this should be ready for another look

PR #129, opened 2018-09-14
Add support for addons and the switch component

jwbennet:
Adds support for addon components which are kept in a separate submodule.  The instructions for this component include details on how to leverage the component since it requires an additional NPM module and a different sort of import.

### DIFF
--- a/addons/index.js
+++ b/addons/index.js
@@ -1,0 +1,1 @@
+module.exports = require('../build/dist/components/Addons');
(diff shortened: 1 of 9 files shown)


jwbennet:
Yeah, the on/off text needs to be static in the current iteration since the switch will not render properly with any other text

levimcg:
Just a heads up, we can adjust the styling of the switch so that can accept other combos for the labeling (`On`/`Off`). `Yes`/`No` is one I could think of, but what I mostly just don't want to encourage folks to put super long strings in there and it become unrecognizable as a switch. If that makes sense.

PR #132, opened 2018-09-15
fix(Collapse): Fix a Typescript issue with the collapse component

jwbennet:


src/components/Addons/Collapse/Collapse.tsx
```diff
@@ -13,7 +13,7 @@ interface CollapseProps {
   /**
    * The component to use for the title element.  Defaults to "div".
    */
-  TitleComponent: string;
+  TitleComponent?: string;
```
jhoerr: I neglected to review the Collapse PR and so didn't see this before now. I think the TitleComponent should be an string union of acceptable options. The use of free-form text means the component could break if I misunderstand the intent of the property and type something that isn't allowed.

jwbennet: If this was enumerated it would need to list out every single possible HTML element which I think would be overkill.

jhoerr: It seems like we can constrain it a little more than that? The Rivet docs say: 
```
The button that toggles the collapsible is wrapped in a heading element in the examples that follow. Any heading level can be used based on the heading structure of your app. Or, if a heading element doesn’t make sense semantically it can be replaced with a generic <div> element.
```
And it seems like some valid elements, such as `input`/`hr`/`br`, wouldn't make sense there. 

As a straw man, how about: `div`|`h1`|`h2`|`h3`|`h4`|`h5`|`h6`?

jwbennet: I suppose we can start there, so I'll make the change.  Just FYI, react-bootstrap constrains a similar property as a `string` though: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-bootstrap/lib/PanelHeading.d.ts#L6

jhoerr:
Looks great, thank you!

src/components/Addons/Collapse/Collapse.tsx
jhoerr: I see your point, but I have to admin that implementation doesn't make much sense to me. They have two properties named `_class`, where one represents a class name and the other represents an element name. To me it's a bit of a smell.

PR #135, opened 2018-09-15
Fixes two display issues with the header identity menu in the drawer

jwbennet:
This corrects two display issues in the header identity menu when it is rendered in a drawer:
* If you did not have a logout action handler there would be an extra right border which was not necessary
* If you have an identity menu with no children it did not have a consistent bottom border like the other elements.  This is a CSS bug in Rivet but is being fixed here for the time being.


jwbennet:
For clarity, before: 
image

After:
image

PR #136, opened 2018-09-15
Remove logic to make button disabled if no onClick provider is specified

jwbennet:
This also cleans up some indentation issues in the Pagination component


jhoerr:
I think this is the right thing to do for the Button. The current behavior goes against my intuition.

PR #137, opened 2018-09-15
Update develop from master before merging vnext

jwbennet:


jwbennet:
Proceeding with this merge since it is uninteresting.

PR #139, opened 2018-09-25
adds table variants

jolamar:


src/components/Table/Table.tsx
```diff
@@ -3,17 +3,19 @@ import * as React from 'react';
 import * as Rivet from '../util/Rivet';
 
 interface TableProps {
-    /** 
+    /**
      * Optional Rivet style which changes the look and feel of the rendered table.
      */
-    variant?: 'stripes' | 'plain'
+    variant?: 'stripes' | 'plain' | 'compact' | 'cells'
```
burnumd: Are these variants actually mutually exclusive? From the documentation, it sounds like either could be applied to a striped or plain table (or alongside each other).

src/components/Table/Table.tsx
```diff
@@ -3,17 +3,19 @@ import * as React from 'react';
...
 }
 
-const Table : React.SFC<TableProps & React.HTMLAttributes<HTMLTableElement>> = 
+const Table : React.SFC<TableProps & React.HTMLAttributes<HTMLTableElement>> =
 ({ children, className, variant, ...attrs }) => {
     const classes = classNames({
         ['rvt-table-plain']: variant === 'plain',
         ['rvt-table-stripes']: variant === 'stripes',
+        ['rvt-table-compact']: variant === 'compact',
```
jwbennet: You might change classes to the following:
```
const classes = classNames({
  [`rvt-table-${variant}`]: variant
});
```

I think having plain and stripes separate the way they are is a result of me copying a component where those were both boolean properties and I never merged them together correctly.

src/components/Table/Table.tsx
levimcg: @burnumd That's a good call. Yes, they design so that they can be applied alongside each other if needed.

jwbennet:
As a general comment, you may also modify the title of this pull request since it appears to be a catch-all for Rivet 1.1 updates.

src/components/Badge/Badge.tsx
```diff
@@ -11,7 +11,7 @@ interface BadgeProps {
     /**
      * Optional Rivet style: an action/error/success/warning badge.
      */
-    variant?: 'action' | 'error' | 'success' | 'warning';
+    variant?: 'action' | 'error' | 'danger' | 'success' | 'warning';
```
jwbennet: I think we should go ahead and remove the `error` variant as long as we mention it in the commit message for the release notes.

src/components/Alert/common.ts
```diff
@@ -5,7 +5,7 @@ export interface StatefulAlertProps {
      * Rivet alert styling. 
      * @see https://rivet.uits.iu.edu/components/overlays/alerts
      */ 
-    variant: "error" | "info" | "message" | "success",
+    variant: "error" | "danger" | "info" | "message" | "success",
```
jwbennet: I think we should go ahead and remove the `error` variant as long as we mention it in the commit message for the release notes.

src/components/Alert/common.ts
```diff
@@ -17,7 +17,7 @@ export interface StatefulAlertProps {
 }
 
 export const statefulPropTypes = {
-    variant: PropTypes.oneOf(['error', 'info', 'message', 'success']).isRequired,
+    variant: PropTypes.oneOf(['error', 'danger', 'info', 'message', 'success']).isRequired,
```
jwbennet: I think we should go ahead and remove the `error` variant as long as we mention it in the commit message for the release notes.

jolamar:
New PR up at #141 for 1.1.0 release

PR #141, opened 2018-09-26
Update Rivet version 1.1.0

jolamar:
- Adds compact and cell table variants
- Adds lr and tb edges
- Adds new typescale aliases
- Consistent color naming: info, success, warning, danger

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "rivet-react",
-  "version": "0.4.0-alpha",
+  "version": "0.4.1",
   "description": "Rivet React components",
   "author": "",
   "license": "MIT",
@@ -42,7 +42,7 @@
     "react-styleguidist": "7.2.5",
     "rivet-collapsible": "0.1.1-alpha",
     "rivet-switch": "0.1.1-alpha",
-    "rivet-uits": "1.1.0-alpha.2",
+    "rivet-uits": "1.1.0",
     "semantic-release": "15.9.12",
     "typescript": "2.9.2"
   },
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "rivet-react",
-  "version": "0.4.0",
+  "version": "0.4.0-alpha",
   "description": "Rivet React components",
   "author": "",
   "license": "MIT",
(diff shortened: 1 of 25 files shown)


jolamar:
I need some help with the table variants which are not mutually exclusive.. @anyone Here are the possibilities of table variations and combinations of variants (@levimcg plz confirm with a 👍)
- plain
- stripes
- compact
- cells
- plain + compact
- stripes + compact
- stripes + cells
- stripes + cells + compact
- cells + compact

jwbennet:
@jolamar you might make sure to merge this with the "Merge pull request" option instead of something like a squash and merge so it preserves all of the good commit messages after this is merged in.

jolamar:
@jwbennet I may need to ammend one of the commits which I think is a breaking change (by removing the 'error' variant). https://github.com/indiana-university/rivet-react/pull/141/commits/8e30ad4782cf33dd15e9f9371158e9c1fb70db8f#diff-9f50914a694c3edd08581b97810b997dL8

jwbennet:
@jolamar it sounds to me like having the `compact` and `cells` as separate boolean properties might be best.  If they are present then add the additional table classes.  For example:

```jsx
<Table variant="stripes" compact>
...
</Table>
```

jwbennet:
If you can amend that commit message easily it might be worthwhile, but if it seems troublesome it will probably be okay.  Since there is a different `feat` commit it will still increment the minor version number correctly to indicate that there are breaking changes.

levimcg:
@jolamar Those combos on the table variants all look good to me. I think the incompatible combos would be:

- plain + cells
- plain + stripes

jolamar:
Would love to get some 👀 on my last commit https://github.com/indiana-university/rivet-react/pull/141/commits/c6085e7894c334343d5a049747c5b434366ffb62

burnumd:
@levimcg @jolamar I'm not sure that `plain` + `cells` are entirely incompatible. I'd expect it to give you a table with bordered cells but a non-shaded header row.

burnumd:
@levimcg @jolamar In practice, with vanilla HTML+Rivet CSS, it seems to lead to a case where you just get vertical borders: https://codepen.io/anon/pen/YOmozj

jolamar:
@burnumd I've created an issue on Rivet for this https://github.iu.edu/UITS/rivet-source/issues/372 If it makes it into Rivet we can adjust Rivet React later

jwbennet:
I think the table changes look good.

levimcg:
@jolamar @burnumd I kinda feel lik this might be a documentation issue. In my mind I would expect the `plain` variant to always remove any additional decoration from the table (borders, background, etc.).

Maybe we could just be more clear about that in the documentation on the Rivet site. That's my bad.

Issue #127, opened 2018-09-13
Tab accessibility support


jwbennet:
The first pass of tabs support will not include keyboard support for accessibility.  We will need to add that to have parity with Rivet core.  See https://github.iu.edu/UITS/rivet-source/blob/develop/src/js/components/tabs.js.

PR #157, opened 2018-10-15
Added expected keyboard navigation support to Tabs component

levimcg:
This PR adds keyboard navigation support to the the Tabs component and should address #127.

Focus management is a _very_ DOM-specific thing, so I'd be interested to talk about how best to handle that in the Rivet React library in a consistent way moving forward. I'm not aware of a way to handle focus expectations in a data-driven/state-specific way, so I've relied heavily on the `ref` React API here.

This is my first foray into Typescript so feedback it definitely welcome :)

I've written tests that pass for me locally, but again, if anyone has ideas about how to improve them (they are a bit repetitive), I'd love some feedback.

### DIFF
--- a/src/components/Tabs/Tabs.test.tsx
+++ b/src/components/Tabs/Tabs.test.tsx
@@ -54,4 +54,120 @@ describe('<Tabs />', () => {
     });
   });
 
+  describe('Tab keyboard interactions', () => {
+    let mock;
+
+    beforeEach(() => {
+      mock = mount(
+        <Tabs>
+          <Tab title="Tab one" id="t-one">
+            Tab one content
+          </Tab>
+          <Tab title="Tab two" id="t-two">
+            Tab two content
+          </Tab>
+          <Tab title="Tab three" id="t-three">
+            Tab two content
+          </Tab>
+        </Tabs>
+      );
+    })
+
+    it('Should focus the first tab by default', () => {
+      mock
+      .find('button#t-one-tab')
+      .simulate('click');
+
+      const focusedElement = document.activeElement;
+
+      expect(mock
+        .find('button#t-one-tab')
+        .matchesElement(focusedElement));
+    });
+
+    it('Should focus the next tab when the right arrow key is pressed', () => {
+      mock
+        .find('button#t-one-tab')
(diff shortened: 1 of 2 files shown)


src/components/Tabs/Tabs.test.tsx
```diff
@@ -54,4 +54,120 @@ describe('<Tabs />', () => {
     });
   });
 
+  describe('Tab keyboard interactions', () => {
+    let mock;
+
+    beforeEach(() => {
+      mock = mount(
```
athill: While I'm not sure of the best term for this, I don't think `mock` is accurate and is potentially misleading. We use `cut` in other tests, which I honestly don't understand, but we should probably use the same name everywhere.

jhoerr:
“cut” means “component under test.” It’s derived from “sut” (“system under test”). Let’s change that nomenclature if it’s confusing. I agree that it’s not a “mock”, though. :)

--
John Hoerr (mobile)
503.360.6581

athill:
@jhoerr Mystery solved (as far as `cut`)! Not sure of a better term offhand. We've used `output` before, but not thrilled with that either.

levimcg:
Huh, I also didn't understand `cut`, good to know. Sounds tougher too. 🔪 

Looks like the code coverage dropped a bit. Not sure how to fix that. Are there specific tests that I'm missing?

jhoerr:
@levimcg It looks like there might be a line or two of code not covered by tests. You can run `yarn test --coverage` and it should show you the component line number(s) that are missing coverage.

levimcg:
Ok, I think I'm done with this. All the checks are passing, code coverage is back at 100%.

jolamar:
screen shot 2018-10-17 at 4 10 40 pm
Testing to make sure it passes these accessibility requirements. Then I'll approve.

jhoerr:
@jolamar Could I impose on you freeze that version of webpack?

levimcg:
@jolamar I’ll be offline for PTO the next couple of days and the away at a conference the first part of next week. Would you be able to take this one from where and finish it up once everyone is done reviewing it? Thanks! 🙌

PR #173, opened 2019-01-30
Update license to BSD 3-clause

jwbennet:
Update the license to BSD 3-clause, add headers to each source file and updated contribution guidelines to explain how to apply the license headers appropriately.


src/components/Addons/Collapse/Collapse.test.tsx
```diff
@@ -1,3 +1,7 @@
+/*
+Copyright (C) 2018 The Trustees of Indiana University
```
athill: Should this be 2019?

athill: Nevermind, from http://opensource.iu.edu/: `[year of first publication]`

PR #205, opened 2019-08-14
Override toggle dropdown click inside

jkneal:


src/components/Dropdown/Dropdown.md
```diff
@@ -31,3 +31,16 @@ By default the dropdown menu is vertically aligned with the left edge of the dro
     <a href="#">Item four</a>
 </Dropdown>
 ```
+
+### Toggle On Click Inside Modifier
+
+By default the dropdown menu will no close when an item within the dropdown is clicked. Set the 'toggleDropdownOnClickInside' option to true to specify the dropdown should be closed in this case.
```
jwbennet: Minor typo here.  I think it should be `not` instead of `no`.

src/components/Dropdown/Dropdown.md
```diff
@@ -31,3 +31,16 @@ By default the dropdown menu is vertically aligned with the left edge of the dro
...
+
+```jsx
+<Dropdown toggleDropdownOnClickInside={true}>
```
jwbennet: Since this is a boolean property you could change this example to remove the `={true}` part.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -59,7 +61,7 @@ export class Dropdown extends React.PureComponent<DropdownProps & React.HTMLAttr
     }
 
     public render() {
-        const { align, children, className, label, menuClass, ...attrs } = this.props;
+        const { align, children, className, label, menuClass, toggleDropdownOnClickInside, ...attrs } = this.props;
```
jwbennet: Might consider being explicit and set `toggleDropdownOnClickInside = false` just so it is clearer on the default behavior.

PR #206, opened 2019-08-16
Feature/toggle dropdown click inside

jkneal:
This fixes a problem with the feature in that the onClick for dropdown children was not getting invoked because the click event was getting captured by the DropdownEvent listener

### DIFF
--- a/src/components/Dropdown/Dropdown.md
+++ b/src/components/Dropdown/Dropdown.md
@@ -34,13 +34,11 @@ By default the dropdown menu is vertically aligned with the left edge of the dro
 
 ### Toggle On Click Inside Modifier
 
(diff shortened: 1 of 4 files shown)


src/components/Dropdown/Dropdown.md
```diff
@@ -38,7 +38,7 @@ By default the dropdown menu will not close when an item within the dropdown is
 
 ```jsx
 <Dropdown label="Toggle On Click" toggleDropdownOnClickInside>
-    <Button variant="plain">Item one</Button>
-    <Button variant="plain">Item two</Button>
+    <Button onClick={() => alert('Clicked Item one')}>Item one</Button>
+    <Button onClick={() => alert('Clicked Item one')}>Item two</Button>
```
burnumd: Did you mean for this alert to read `'Clicked Item two'`?

jkneal: Yes, good catch!

PR #219, opened 2020-02-04
Upgrade dependencies and retain TypeScript

jhoerr:
* Update to react 16.12
* Update to typescript 3.7.5
* Update all other packages to latest.

There unfortunately remain some `yarn audit` findings, but they are limited to dev dependencies (`codecov.io`, `semantic-release`) and won't ship with the package.


jhoerr:
A `yarn audit --groups dependencies` comes up clean. There remain vulnerable packages for `codecov.io` and `semantic-release`, but these are dev dependencies.

burnumd:
Can you change the `^` dependencies to exact versions so we have consistent dependency versioning?

burnumd:
Also, this PR should be against `master` not `develop`.

jhoerr:
> Can you change the ^ dependencies to exact versions so we have consistent dependency versioning?

Will do.

> Also, this PR should be against master not develop.

We should get rid of the `develop` branch if we're not doing git flow anymore. It is currently out of sync with master, and is now a liability.

burnumd:
> We should get rid of the develop branch 
Agreed. I created #220 to track it

jhoerr:
I turned off that stupid codecov rule, whatever it's tracking.

jhoerr:
Resolved a few minor merge conflicts with `master`.

jhoerr:
@nublikasa @burnumd @wsams @jwbennet @athill @kennethj 

I think this is ready for review. I'm happy to answer any questions or help with evaluation. Release the hounds!

package.json
```diff
@@ -16,37 +16,39 @@
...
-    "semantic-release": "15.9.12",
-    "typescript": "2.9.2",
-    "webpack": "3.8.1"
+    "enzyme": "3.11.0",
+    "enzyme-adapter-react-16": "1.15.2",
+    "gh-pages": "2.2.0",
+    "license-check-and-add": "3.0.4",
+    "react-docgen-typescript": "1.16.2",
+    "react-dom": "16.12.0",
+    "react-scripts": "^3.3.1",
```
burnumd: This dependency should use the exact version

```suggestion
    "react-scripts": "3.3.1",
```

src/components/Alert/Alert.tsx
```diff
@@ -3,13 +3,34 @@ Copyright (C) 2018 The Trustees of Indiana University
 SPDX-License-Identifier: BSD-3-Clause
 */
 import * as classNames from 'classnames';
+import * as PropTypes from "prop-types";
```
burnumd: We're using single quotes in imports elsewhere

```suggestion
import * as PropTypes from 'prop-types';
```

src/components/Input/Select.tsx
```diff
@@ -7,7 +7,7 @@ import * as Rivet from '../util/Rivet';
 import { renderInput, TextProps } from './common';
 
 const Select : React.SFC<TextProps & React.SelectHTMLAttributes<HTMLSelectElement>> = 
-    renderInput((attrs) => <select {...attrs}/>);
+    renderInput("select");
```
burnumd: This should use single quotes to be consistent

```suggestion
    renderInput('select');
```

src/components/Input/Textarea.tsx
```diff
@@ -7,7 +7,7 @@ import * as Rivet from '../util/Rivet';
 import { renderInput, TextProps } from './common';
 
 const Textarea :  React.SFC<TextProps & React.TextareaHTMLAttributes<HTMLTextAreaElement>> = 
-    renderInput((attrs) => <textarea {...attrs}/>);
+    renderInput("textarea");
```
burnumd: This should use single quotes to be consistent

```suggestion
    renderInput('textarea');
```

src/components/List/List.tsx
```diff
@@ -26,7 +26,7 @@ const asListItem = (child : React.ReactChild) =>
     : child;
 
 /** Ensure all children are <li> elements. */
-const asListItems = (children?: React.ReactNode) => 
+const asListItems = (children?: any) =>
```
burnumd: Why was this restriction relaxed and doesn't use `React.ReactChild`?

src/components/Pagination/Pagination.tsx
```diff
@@ -26,16 +26,16 @@ interface PaginationItemProps {
...
+    const classes = classNames.default({
       ['rvt-pagination']: true,
       [`rvt-pagination--${align}`]: align,
       [`rvt-pagination--${size}`]: size
     });
-    const wrappedChildren = React.Children.map(children, (child: React.ReactElement<any>) => {
-      const childProps = child.props as PaginationItemProps;
-      const childClasses = classNames({
+    const wrappedChildren = React.Children.map(children as JSX.Element[], (child: React.ReactChild) => {
+      const childProps = child.hasOwnProperty("props") ? child['props'] as PaginationItemProps : {};
```
burnumd:
```suggestion
      const childProps = child.hasOwnProperty('props') ? child['props'] as PaginationItemProps : {};
```

src/components/util/Rivet.tsx
```diff
@@ -149,9 +149,9 @@ const parseRivetHidden = (hide?: Hidden) => ({
  */
 export type ComponentDecorator<T extends Props> = ( props: T ) => string;
 
-export const rivetize = <T extends React.HTMLAttributes<HTMLElement>>(Component: React.ComponentType<T>): React.SFC<T & Props> =>
-({ className, border, display, hide, margin, padding, typescale, ...attrs }: Props & React.HTMLAttributes<HTMLElement>) => (
-        <Component className={classnames(
+export const rivetize = <T extends React.HTMLAttributes<HTMLElement>>(Component: React.ComponentType<any>): React.FC<T & Props> =>
+    ({ className, border, display, hide, margin, padding, typescale, ...attrs }: T & Props) => (
```
burnumd: Is the default spacing for `.tsx` files 2 spaces or 4?

src/setupTests.ts
```diff
@@ -11,5 +11,5 @@ window.requestAnimationFrame = function (callback) {
   };
 
 import * as Enzyme from 'enzyme';
-import * as Adapter from 'enzyme-adapter-react-16';
+import Adapter from 'enzyme-adapter-react-16';
 Enzyme.configure({ adapter: new Adapter() });
```
burnumd:
```suggestion
Enzyme.configure({ adapter: new Adapter() });

```

jhoerr:
As a general note, perhaps we could consider fixing stylistic and linting stuff on a separate branch? I'd like to keep this focused on design decisions.

jhoerr:
> Can you point to some documentation (or just describe) what changed that prevents us from using common interfaces and prop types to describe, for example, Alerts and Dismissible Alerts using the same object?

We were already repeating ourselves with `InlineAlert`, and the refactored type hierarchy of `Stateless` inheriting from `Stateful` felt counterintuitive and confusing. To me it felt like a constraining, low-value refactor, so I got rid of it.

src/components/List/List.tsx
jhoerr: `React.ReactNode` yielded a compilation error due to weirdness with `React.Children.map` overloads. This is an internal function so I let it go.

src/components/util/Rivet.tsx
jhoerr: 🤷🏻‍♂️. We should develop a strategy for automatic formatting and linting.

src/components/List/List.tsx
jhoerr: I'll take another swing at it.

src/components/Addons/Collapse/Collapse.tsx
```diff
@@ -50,7 +50,7 @@ class Collapse extends React.PureComponent<CollapseProps & React.HTMLAttributes<
 
   public render() {
     const { children, className, defaultClosed, title, TitleComponent = 'div', variant, ...attrs } = this.props;
-    const classes = classNames({
+    const classes = classNames.default({
```
athill: Why was this `.default` added? Is it a TS thing? I'm not seeing it in the `classNames` docs

jhoerr: I think it's a TS thing. This happened following the update of the `react-scripts` and `typescript` dependencies. It's something to do with the way the exports are set up on `classNames` typings. I tried several things and this is the only thing I could figure out that made the compiler happy. JS/TS imports and exports are baffling to me. I'm happy to take a different approach if you can recommend one.

nublikasa: @jhoerr, we fixed this issue earlier by importing classnames as follows: `import classNames from 'classnames';`. Just an FYI

jhoerr: Fixed with @nublikasa's advice.

burnumd:
One last note since it's been a while: don't forget to use Angular commit format when you squash and merge so it gets released properly!

jhoerr:
Thank you, @burnumd! 😅 Do we consider these updates breaking changes, features, or chores?

burnumd:
It would be a `feat:` as far as I can tell. We also don't want to mark it as a chore because that's just for internal changes that don't require a release (we'd probably mark a linting change as a chore). Whether it's a breaking change is something you two (@jhoerr or @nublikasa) would know better. Does it require a migration path, e.g., will users need to update other dependencies or make any changes to their client code to use this version?

jhoerr:
> will users need to update other dependencies or make any changes to their client code to use this version?

It's hard to know for sure since we don't publish betas. The update from React 16.4 -> 16.12  introduced new types, such as `React.FC`. This may require consumers to update their version of React to stay compatible, but that gets deeper into the weeds on module versioning and resolution than I understand.

(Might be a good time to mention that `semantic-release` [can publish betas now](https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/pre-releases.md).)

package.json
```diff
@@ -16,37 +16,39 @@
...
-    "rivet-collapsible": "0.2.10",
-    "rivet-switch": "0.3.1",
-    "rivet-uits": "1.6.0",
-    "semantic-release": "15.9.12",
-    "typescript": "2.9.2",
-    "webpack": "3.8.1"
+    "enzyme": "3.11.0",
+    "enzyme-adapter-react-16": "1.15.2",
+    "gh-pages": "2.2.0",
+    "license-check-and-add": "3.0.4",
```
burnumd: I just noticed this breaks the `license` script. If you want to fix that in a later PR, this dependency shouldn't be updated now.

jhoerr: Reverted to previous version.

jhoerr:
@burnumd @nublikasa @athill @kennethj @wsams 
I've published the updates to a beta package, `1.4.0-beta.1`.  I'd like to get feedback from you on whether this new package introduces problems/conflicts, particularly for TS apps. If you have a moment, please update one of your apps and confirm that it builds/tests/runs as expected. If it does, great. If not, please include some details about problems you encounter.

jhoerr:
I made one small tweak to the TS transpiler options and published it as `1.4.0-beta.2`. Please test with that instead. Thanks!

athill:
I've confirmed things are generally working in one of our apps (non-TS). More testing required, but looks promising!

burnumd:
I was able to build rac-common and RAC with the beta build without making any additional changes, so I don't believe this update causes any breaking changes.

jhoerr:
My own testing of `1.4.0-beta.2` in a TS app (itpeople-app) looks good as well. I'll go ahead and publish this as a feature update. Thanks @burnumd and @athill!

PR #234, opened 2020-02-07
chore(eslint): Clean up lint warnings

nublikasa:
- Clean up ESLINT warnings
- Add scripts to run linting and fix linting in package.json


.eslintignore
```diff
@@ -0,0 +1,2 @@
+/src/**/*.test.*
```
burnumd: Why do we want to ignore linting on tests?

nublikasa: Good suggestion. I've fixed the lint warnings in test.

PR #240, opened 2020-06-25
feat(Header): Allow custom Header URL

athill:


src/components/Header/Header.test.tsx
```diff
@@ -7,10 +7,11 @@ import * as React from 'react';
 import Header from './Header'
 import HeaderIdentity from './HeaderIdentity'
 import HeaderNavigation from './HeaderNavigation'
+import { isMainThread } from 'worker_threads';
```
burnumd, line 10: What is this import for? I don't see it being used.

athill: I have no idea where that came from. Removing. Thanks.

athill: Actually, I do know. {{vscode}} incorrectly auto-completed something and then auto-imported it. Thanks for catching.

src/components/Header/Header.tsx
```diff
@@ -11,6 +11,11 @@ import Identity from './HeaderIdentity';
 import Navigation from './HeaderNavigation';
 
 interface HeaderProps {
+    /**
+     * The URL that the link in the upper left of the header will point to.
+     * @see https://rivet.uits.iu.edu/components/navigation/header/
+     */
+    headerUrl: string;
```
jwbennet, line 18: My two cents is that I'd expect this to be named `href` to fall in line with the other components which have similar properties (button, etc).

athill: Agreed. Fixing.

src/components/Header/Header.md
```diff
@@ -1,29 +1,33 @@
...
+default view. The base header is the bare minimum version of the header that must
 be included in your application.
 
 View the [Rivet documentation for Header](https://rivet.uits.iu.edu/components/navigation/header/).
 
-
 ```jsx
 <Header title="Application Title" />
 ```
+You can also overrice the default "`/`" URL for the link:
```
burnumd, line 11: Typo
```suggestion
You can also override the default "`/`" URL for the link:
```

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "baseUrl": "src",
     "outDir": "build/dist",
-    "module": "commonjs",
+    "module": "esnext",
```
burnumd, line 5: Why did this change? It wasn't originally part of the PR.

athill: I don't know, but this is the second time I've reverted it. Weird. Thanks for catching.

src/components/Header/Header.md
athill: Fixed. Thanks.

Issue #254, opened 2021-02-19
Add property to Dropdown component to not render role=menu


mrw-iu:
We use the Dropdown component in two of our tools to display/hide options for filtering the current view.  After going through an accessibility review, the recommendation is that we should not be using the role=menu attribute for our use case.  It would be helpful if the Dropdown component had an additional property to optionally render the Dropdown without this accessibility role.  I will provide a pull request.

PR #255, opened 2021-02-19
Add property to Dropdown component to not render role=menu

mrw-iu:
We use the Dropdown component in two of our tools to display/hide options for filtering the current view (LMS tools for Canvas). After going through an accessibility review, the recommendation is that we should not be using the role=menu attribute for our use case. It would be helpful if the Dropdown component had an additional property to optionally render the Dropdown without this accessibility role. I will provide a pull request.

https://github.com/indiana-university/rivet-react/issues/254

### DIFF
--- a/src/components/Dropdown/Dropdown.test.jsx
+++ b/src/components/Dropdown/Dropdown.test.jsx
@@ -256,4 +256,46 @@ describe('<Dropdown />', () => {
             expect(UnwrappedDropdown.prototype.toggleDropdown).toHaveBeenCalledTimes(1);
         });    
     });
+
+    describe('Menu Role Behavior', () => {
+
+            it('role=menu should render by default', () => {
+                let cut = mount(<Dropdown>
+                                <a href="#">Hello, world!</a>
+                            </Dropdown>, { attachTo: root });
+                cut.find('button').simulate("click");
+
+                expect(cut.find( 'div.rvt-dropdown__menu[role="menu"]' )).toHaveLength(1);
+
+            });
+
+            it('excludeMenuRole=true should not render the role=menu', () => {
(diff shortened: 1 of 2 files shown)


burnumd:
Hello @mrw-iu,

Thanks for this contribution! Can you give more information on the guidance you received about removing the `menu` role from the dropdown component? As I understand it, ARIA roles describe how users can interact with an element, and in the case of `menu`, it [determines several keyboard navigation features](https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus#Success_Criteria_2.1.1_Keyboard) which are going to continue to be present and should therefore notify a screen reader that this element is a `menu`.

burnumd (changes requested):
Please [add documentation](https://github.com/indiana-university/rivet-react/blob/master/CONTRIBUTING.md#contribution-requirements) to `Dropdown.md` describing this new option.

mrw-iu:
Thank you for looking at this so quickly.  I talked to our accessibility contact today to see if this change will cause negative effects because of the other accessibility features built in to this component (the keyboard navigation as well as other ARIA), and he is going to do some user testing without this role to see if that will be the case.  So please just put this one on hold for the moment and I will let you know the results.

mrw-iu:
I apologize for the delay in responding.  We still haven't heard back with the additional accessibility testing results.  In the meantime, we decided to just use a different approach for our scenario.  So this request can be closed without merging.  Thank you so much for your help and time on this issue.  Let me know if you would prefer to close it yourself or if you would like me to close it.

burnumd:
@mrw-iu Thanks for the update! I'll go ahead and close out this PR and the associated issue, but thanks again for the contribution and please feel free to create a new one if you get any further guidance.

PR #258, opened 2021-03-19
feat(Footer): Footer copyright shows current year

nublikasa:


src/components/Footer/Footer.test.tsx
```diff
@@ -40,6 +40,23 @@ describe('<Footer />', () => {
...
+            expect(cut.find('li.rvt-footer__aux-item').at(1).text()).toEqual(copyrightText);
+        });
+        it('should render copyright year argument specified', () => {
+            const year = 2020;
+            const copyrightText = "Copyright © " + year + " The Trustees of Indiana University";
+            const cut = mount(<Footer copyrightYear={year} />);
+            expect(cut.find('li.rvt-footer__aux-item').at(1).text()).toEqual(copyrightText);
+        });
+        it('should render copyright with current year when given a non number', () => {
+            const year = "2020 - 2021";
```
burnumd, line 55: Strings in JS are typically single quoted
```suggestion
            const year = '2020 - 2021';
```

burnumd: The `copyrightText` field in all of these tests should also use single quotes.

src/components/Footer/Footer.tsx
```diff
@@ -7,14 +7,25 @@ import * as React from 'react';
 import * as Rivet from '../util/Rivet';
 import Icon from '../util/RivetIcons';
 
+interface FooterProps {
+    /**
+     * The Copyright year that appears in the footer.
+     * @see https://rivet.iu.edu/components/navigation/footer/
+     */
+    copyrightYear?: number;
```
burnumd, line 15: I thought we also wanted to accept arbitrary strings so users could specify a range?

burnumd (changes requested):

burnumd (approved):

PR #291, opened 2022-08-22
feat: EPMDW-432 -  2022 Q3 Dependency updates

athill:


tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "baseUrl": "src",
     "outDir": "build/dist",
-    "module": "commonjs",
+    "module": "esnext",
```
burnumd, line 5: IIRC, this is something that should be set to `commonjs` but VSCode will update it to `esnext` for some reason. Does that sound familiar?

src/setupTests.ts
```diff
@@ -5,7 +5,7 @@ SPDX-License-Identifier: BSD-3-Clause
 // Temporary hack to suppress error
 // https://github.com/facebookincubator/create-react-app/issues/3199
 import * as Enzyme from 'enzyme';
-import Adapter from 'enzyme-adapter-react-16';
+import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
```
burnumd, line 8: Can you put a note in a comment above this import explaining why we're using a 3rd-party adapter (maybe linking to the Enzyme deprecation) so we know to keep an eye on its continued development?

burnumd (changes requested):

tsconfig.json
athill: It does. I'll revert it. Thanks.

src/setupTests.ts
athill: Yes. Good call.

tsconfig.json
athill: Fixed. Thanks.

src/setupTests.ts
athill: Fixed. Thanks.

burnumd (approved):

PR #309, opened 2022-11-11
Rivet2 buttons cleanup and tests

pcberg:


pcberg:
Prettier seems to have made some formatting on the documentation .md files. Do we like this?

Screen Shot 2022-11-11 at 17 59 09

burnumd:
> Prettier seems to have made some formatting on the documentation .md files. Do we like this?

Yeah, let's discuss this Thursday. It doesn't really match up with the Rivet docs, I don't think.

src/components/Button/SegmentedButton.md
```diff
@@ -3,20 +3,61 @@ Use the segmented button to display a group of related controls in a single line
 View the [Rivet documentation for Segmented Buttons](https://rivet.uits.iu.edu/components/forms/buttons-segmented/).
 
 ### Default Style
+
 ```jsx
 <SegmentedButton>
-    <Button onClick={() => { console.log('Click 1!') }}>Primary One</Button>
-    <Button onClick={() => { console.log('Click 2!') }}>Primary Two</Button>
-    <Button onClick={() => { console.log('Click 3!') }}>Primary Three</Button>
+  <Button
```
burnumd, line 9:
```suggestion
<!-- prettier-ignore-start -->
    <Button
```

src/components/Button/SegmentedButton.md
```diff
@@ -3,20 +3,61 @@ Use the segmented button to display a group of related controls in a single line
...
+    Primary Two
+  </Button>
+  <Button
+    onClick={() => {
+      console.log("Click 3!");
+    }}
+  >
+    Primary Three
+  </Button>
 </SegmentedButton>
```
burnumd, line 30:
```suggestion
<!-- prettier-ignore-end -->
</SegmentedButton>
```

pcberg: @burnumd thanks, addressed! Looks better now

src/components/Button/Button.md
```diff
@@ -3,6 +3,9 @@ Buttons are used to signal actions, submit forms, or trigger new content within
 View the [Rivet documentation for Buttons](https://rivet.uits.iu.edu/components/forms/buttons/).
 
 ### Button Examples
+
+<!-- prettier-ignore-start -->
```
burnumd, line 7: There should also be a corresponding `<!-- prettier-ignore-end -->` for each start comment.

burnumd (changes requested):

ri-pandey (commented):
You might want to consider not using class/id-based selectors, and instead use the kind of queries described at https://testing-library.com/docs/queries/about. I think that would be more along the lines of the recommendations of testing-library. If you do that, you would also have to replace some attribute selectors (like `classList.contains`) with `toHaveAttribute`, etc. (more at https://github.com/testing-library/jest-dom). I noted one such place in one of the tests.

src/components/Button/Button.md
pcberg: Woops! Thanks, revised. I hope it worked better this time...

src/components/Button/Button.test.jsx
```diff
@@ -0,0 +1,157 @@
...
+import "@testing-library/jest-dom";
+import user from "@testing-library/user-event";
+import React from "react";
+import Button from "./Button";
+
+describe("<Button />", () => {
+  describe("Rendering and text", () => {
+    it("should render without throwing an error", () => {
+      const { container } = render(<Button />);
+      expect(container.querySelectorAll("button.rvt-button")).toHaveLength(1);
```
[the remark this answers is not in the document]
pcberg: Do you have examples of these scenarios in Dropdown.test.jsx?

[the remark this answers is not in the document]
ri-pandey: Yes, [here](https://github.com/indiana-university/rivet-react/blob/feature/2-0-0-build-dropdown/src/components/Dropdown/Dropdown.test.jsx#L17). Assuming there is only one <button> element in your test, you can select it by role "button".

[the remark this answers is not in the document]
pcberg: Thanks Rishi, I'll take a look. Yes, I replaced some instances with getByRole but I couldn't figure out all scenarios. I'll give it one more pass. One followup question: So it's safe to assume that
`it("should render without throwing an error"`
in [SegmentedButton.test.jsx](https://github.com/indiana-university/rivet-react/pull/309/files#diff-d462e47710842515b9cec0196f0f1009e40190715039615acca561ac3fa88dda) I can't do with getByRole, right? I was actually wondering if that's a useful test but I got it from V1 and didn't feel comfortable just killing it.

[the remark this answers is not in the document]
ri-pandey: You could replace `container.querySelectorAll("div")` with `getByText("One")` for that. See [ByText](https://testing-library.com/docs/queries/bytext). The `container.querySelectorAll("button")` can be replaced with `getAllByRole("button")`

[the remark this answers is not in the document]
pcberg: @ri-pandey thanks so much for your feedback! I was able to replace all container queries with user-interaction testing. Looks much better! Take a look and see if you spot anything else. I did add TestUtils per your example, I think that's the biggest thing I was missing for some of my queries.

src/components/Button/Button.test.jsx
```diff
@@ -0,0 +1,112 @@
...
+      expect(screen.getByRole("button", {})).toHaveClass(
+        "rvt-button--plain-secondary"
+      );
+      expect(screen.getByRole("button", {})).toHaveClass("rvt-button--small");
+    });
+
+    it("should apply appropriate navigation classes", () => {
+      render(<Button variant="navigation" />);
+      expect(screen.getByRole("button", {})).not.toHaveClass("rvt-button");
+      expect(screen.getByRole("button", {})).toHaveClass(
```
burnumd, line 95: A small thing, but this could be changed to 

```suggestion
      const button = screen.getByRole("button", {});
      expect(button).not.toHaveClass("rvt-button");
      expect(button).toHaveClass(
```

burnumd: There are a couple other locations where we have a `screen.getByRole("button", {})` twice in one test that it would be nice to clean up.

burnumd (changes requested):

burnumd (approved):

PR #310, opened 2022-11-18
Feature/2 0 0 build dropdown

ri-pandey:


src/components/Alert/index.jsx
```diff
@@ -2,5 +2,5 @@
 Copyright (C) 2018 The Trustees of Indiana University
 SPDX-License-Identifier: BSD-3-Clause
 */
-export { default as Alert } from "./Alert";
-export { default as DismissibleAlert } from "./DismissibleAlert";
+export { default as Alert } from "./Alert.jsx";
+export { default as DismissibleAlert } from "./DismissibleAlert.jsx";
```
burnumd, line 6: Is this something prettier did? We don't have file extensions on other internal imports/exports.

src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,174 @@
...
+  const handleClickOutside = (event) => {
+    if (event && shouldToggleDropdown(event)) {
+      toggleDropdown(event);
+
+      if (isEscapeKeyPress(event) && toggleButton.current) {
+        // If the user pressed the escape key and we have a current reference to the dropdown toggle button put focus back on it
+        toggleButton.current.focus();
+      }
+    }
+  };
```
burnumd, line 42: Can these be moved outside the component declaration so they aren't reinitialized every time the component rerenders? If not, they should go after any `use*` hooks by convention.

src/components/Dropdown/Dropdown.md
```diff
@@ -0,0 +1,50 @@
+Use the dropdown component to create a list of menu options that can be toggled with a button element.
+
+View the [Rivet documentation for Dropdowns](https://rivet.iu.edu/components/dropdown/).
+
+### Dropdown Example
+
+Use the `label` property to set the text on the dropdown button.
```
burnumd, line 7: We might either tweak the wording here or add another example to show that React nodes can be used for labels as well.

burnumd (changes requested):

src/components/Alert/index.jsx
ri-pandey: No, but for some reason styleguidist wouldn't run this without the suffix. I checked and it isn't happening now. Weird. Will remove it.

src/components/Dropdown/Dropdown.jsx
ri-pandey: I moved the hooks at the beginning of the method.

src/components/Dropdown/Dropdown.md
ri-pandey: Added an example.

burnumd (approved):

PR #319, opened 2022-12-10
Build Input (Input, Textarea, Select) Components

pcberg:
1. The spacing and width of these components looks different then 1.0 because of changes to various classNames (example: rvt-container-sm rvt-p-all-xxl-lg-up rvt-m-bottom-md rvt-p-all-lg).
a. Examples before:
Screenshot 2022-12-09 at 19 01 33
Screenshot 2022-12-09 at 19 03 28
Screenshot 2022-12-09 at 19 02 11
b. Examples after:
Screenshot 2022-12-09 at 19 01 10
Screenshot 2022-12-09 at 19 03 59
Screenshot 2022-12-09 at 19 02 40
3. In order to fix this warning (also for Select & Textarea):
`Warning: src/components/Input/Input.jsx matches a pattern defined in “components” or “sections” options in your style guide config but doesn’t export a component.`
I changed renderInput in common.jsx from a higher order function to a regular one and added an empty component (`<>{renderInput("input", props)}</>`). That gets rid of the warning at the expense of slightly wordy code

I am interested in feedback if anybody spots flaws / improvements on the above. Thanks!

### DIFF
--- a/src/components/Input/common.jsx
+++ b/src/components/Input/common.jsx
@@ -1,0 +1,86 @@
+/*
+Copyright (C) 2018 The Trustees of Indiana University
+SPDX-License-Identifier: BSD-3-Clause
+*/
+import * as React from "react";
+
+import InlineAlert from "../Alert/InlineAlert";
+import * as Rivet from "../util/Rivet";
+import classNames from "classnames";
+import * as PropTypes from "prop-types";
+
+const inputClassName = (elementName, variant) => {
+  const validationVariant = variant ? `rvt-validation-${variant}` : "";
+
+  switch (elementName) {
+    case "input":
+      return classNames("rvt-text-input", validationVariant);
+    default:
+      return classNames(`rvt-${elementName}`, validationVariant);
+  }
+};
+
+const noteFragment = (id, variant, note) =>
+  variant ? (
+    <InlineAlert id={id} variant={variant}>
+      {note}
+    </InlineAlert>
+  ) : (
+    <small id={id} className="rvt-display-block">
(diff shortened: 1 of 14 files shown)


src/components/Input/common.jsx
```diff
@@ -0,0 +1,87 @@
...
+  };
+  return (
+    <div
+      className={classNames(
+        "rvt-container-sm",
+        "rvt-p-all-lg",
+        "rvt-p-all-xxl-lg-up",
+        className
+      )}
+    >
```
ri-pandey, line 73: Why are we wrapping the return inside this div? I think your extra spacing/centre-alignment issues in the docs are coming from this. If you remove this div, your components should not have any extra spacing.

ri-pandey (changes requested):

src/components/Input/common.jsx
pcberg: This makes so much more sense. I copied the styleguide markup. Woops! I also understand now why the original markup documentation sample had margins. I added those back. I do still have the div because we use that as a pass through for custom className. I hope that makes sense, I got it from 1.0: https://github.com/indiana-university/rivet-react/blob/71e2da570662f15829989c2f9bce7562bd0c18c9/src/components/Input/common.tsx#L44

It does look like `font-weight: 700;` was removed in rivet 2.0 from rvt-label. But unless I'm crossing more wires it looks like that was an intentional design choice.

Thanks so much! Let me know what you think on the last revision. It looks so much better now.

ri-pandey (approved):

PR #338, opened 2022-12-28
feat: 2.0.0 EmptyState component

athill:


src/components/EmptyState/EmptyState.md
```diff
@@ -0,0 +1,54 @@
...
+```
+<!-- prettier-ignore-end -->
+
+### EmptyState with Actions
+
+<!-- prettier-ignore-start -->
+```jsx
+<EmptyState>
+    <EmptyState.Content>
+    <p>There's nothing here yet.</p>
```
burnumd, line 25: This line should be indented

burnumd (changes requested):

src/components/EmptyState/EmptyState.md
```diff
@@ -0,0 +1,54 @@
...
+
+View the [Rivet documentation for EmptyState](https://rivet.iu.edu/components/empty-state/).
+
+### EmptyState examples
+
+<!-- prettier-ignore-start -->
+```jsx
+<EmptyState>
+    <EmptyState.Content>
+    <p>There's nothing here yet.</p>
```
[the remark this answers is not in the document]
athill: Fixed. Thanks!

src/components/EmptyState/EmptyState.md
athill: Fixed. Thanks!

src/components/EmptyState/EmptyState.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+  let subComponents = subComponentList.map((key) => {
+    return React.Children.map(children, (child) =>
+      child.type.name === key ? child : null
+    );
+  });
+
+  return (
+    <div className={classPrefix} {...attrs}>
+      {subComponents.map((component) => component)}
```
ri-pandey, line 25: Can this be replaced by `{subComponents}`, instead of the map?

ri-pandey:
@athill I noticed that our approach of separating child components into their own modules only works if the components aren't wrapped in Rivet.rivetize. I wanted to give you a heads up while I investigate this.

ri-pandey:
@athill I managed to fix the root issue on my branch feature/2-0-0-header. You can either rebase with my branch, or wait until I merge my Header branch. Although that won't be for until next week.

athill:
@ri-pandey Let me know when you merge yours, please.

johglove (approved):

src/components/EmptyState/EmptyState.md
```diff
@@ -0,0 +1,54 @@
...
+<!-- prettier-ignore-end -->
+
+### EmptyState with Actions and Icon
+
+<!-- prettier-ignore-start -->
+```jsx
+<EmptyState>
+    <EmptyState.Content>
+        <div className="rvt-p-all-md rvt-m-bottom-md rvt-bg-black-100 rvt-inline-flex rvt-border-radius-circle">
+        <svg width="16" height="16" viewBox="0 0 16 16" aria-hidden="true">
```
basham, line 42: The `svg` should be indented.

src/components/EmptyState/EmptyState.jsx
```diff
@@ -0,0 +1,32 @@
...
+import Content from "./Content";
+
+const classPrefix = "rvt-empty-state";
+
+// https://dev.to/hey_yogini/create-react-subcomponents-in-a-simple-way-5h1f
+
+const EmptyState = ({ children, ...attrs }) => {
+  let subComponentList = Object.keys(EmptyState);
+
+  let subComponents = React.Children.map(children, (child) =>
```
basham, line 16: Restricting the children to only the Actions or Content components seems like an unnecessary restriction, which may cause frustration if the user needs to do something a little different — like even including an extra `div` or something. It should be on the user to use the components in the right way. I recommend to just let the children pass through untouched.

src/components/EmptyState/EmptyState.jsx
```diff
@@ -0,0 +1,24 @@
...
+Copyright (C) 2018 The Trustees of Indiana University
+SPDX-License-Identifier: BSD-3-Clause
+*/
+import * as React from "react";
+import Actions from "./Actions";
+import Content from "./Content";
+
+import * as Rivet from "../util/Rivet";
+
+// https://dev.to/hey_yogini/create-react-subcomponents-in-a-simple-way-5h1f
```
basham, line 11: Remove this icon. It had to do with the approach that was just replaced.

src/components/util/RivetIcons.jsx
```diff
@@ -37,6 +38,15 @@ const IconCore = ({ children, ...attrs }) => (
   </svg>
 );
 
+const icoBell = (attrs) => (
```
basham, line 41: I think this is a typo. It should be `iconBell`, not `icoBell`.

src/components/EmptyState/Actions.jsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+import * as Rivet from "../util/Rivet";
+
+const Actions = ({ children, ...attrs }) => (
+  <div {...attrs} className="rvt-empty-state__actions">
```
basham, line 6: Should `className` merge (rather than replace) classes brought in with `attrs.className`? Is this the pattern done with other components?

src/components/EmptyState/Content.jsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+import * as Rivet from "../util/Rivet";
+
+const Content = ({ children, ...attrs }) => (
+  <div {...attrs} className="rvt-empty-state__content">
```
basham, line 6: Should `className` merge (rather than replace) classes brought in with `attrs.className`? Is this the pattern done with other components?

basham (changes requested):

src/components/util/RivetIcons.jsx
burnumd: Yeah, I agree. The rest of the icons are labeled as `ico*`, though. I think we can change them in another PR.

src/components/EmptyState/Actions.jsx
burnumd: Good question, I'll need to look how we're handling that.

basham (approved):

PR #340, opened 2022-12-28
feat: 2.0.0 RadioButton component

athill:


src/components/RadioButton/RadioButton.md
```diff
@@ -0,0 +1,49 @@
...
+            label="Just a quick note" 
+            description="This is a description for the radio button."
+         />
+    </li>
+  </ul>
+</fieldset>
+```
+<!-- prettier-ignore-end -->
+
+### RadioButton Examples
```
burnumd, line 33: It looks like this is an inline radio button example. You might clarify that in the prose part of the documentation or the heater (otherwise there are just two identical headers).

burnumd (changes requested):

src/components/RadioButton/RadioButton.md
athill: Fixed. Thanks!

burnumd (approved):

PR #341, opened 2022-12-28
feat: 2.0.0 Breadcrumbs component

athill:


burnumd (commented):
Doesn't the index need to export from Breadcrumbs.jsx?

athill:
@burnumd Good catch! Fixed. Thanks!

burnumd (approved):

PR #342, opened 2022-12-29
feat: 2.0.0 File component

athill:


src/components/File/File.jsx
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  componentDidMount() {
+    const form = this.fileInput.current && this.fileInput.current.form;
+    if (form) {
+      form.onreset = (e) => {
+        this.forceUpdate();
+      };
+    }
+  }
```
ri-pandey, line 28: When beginning development for 2.x, we settled on not using lifecycle methods, and replacing them with hooks instead. Is it possible to use hooks here?

athill: Fixed. Thanks.

ri-pandey (approved):

src/components/File/File.jsx
```diff
@@ -0,0 +1,118 @@
...
+  innerRef,
+  label,
+  multiple,
+  onChange,
+  secondary,
+  ...attrs
+}) => {
+  const forceUpdate = useForceUpdate();
+
+  const fileInput = innerRef || React.useRef();
```
burnumd, line 39: Hooks should never be conditional. I'm pretty sure the way of doing what you want is to pass the initial value to `useRef`

```suggestion
  const fileInput = React.useRef(innerRef);
```

src/components/File/File.jsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const form = fileInput.current && fileInput.current.form;
+  if (form) {
```
burnumd, line 42: Codecov is suggesting there should be a test for instances when there is no form (you could mock/spy `useForceUpdate` and check that the resulting function is never invoked.

burnumd (changes requested):

src/components/File/File.jsx
athill: Fixing,. Thanks.

src/components/File/File.jsx
athill: Fixing. Thanks.

pcberg:
@burnumd I made a tweak per https://github.com/indiana-university/rivet-react/pull/342/commits/8a4be60bbb06b7f3afe25d0f765eb5c0420555b4 to get the test in question to pass but I can't say I understand why the await prior to that line isn't working. I followed tips per https://stackoverflow.com/questions/66661163/react-testing-librarys-waitfor-not-working but didn't make traction. Is this "good enough" or does this point to a larger issue we should investigate? Thanks!

burnumd:
@pcberg I think it's "fine for now" if you wouldn't mind filing an issue to investigate why the `waitFor()` behavior isn't working.

pcberg:
@burnumd, I debugged with Rishi a bit and we did some additional cleanup. With respect to waitFor() we think:
1. Since fireEvent is synchronous we don't need waitFor()
2. We need the setTimeout for React to do the extra rendering that we dispatched with forceUpdate

I just pushed an update, do you agree an Issue to investigate is not necessary anymore?

burnumd (approved):

burnumd:
@pcberg I think just adding a comment on that section where we do the timeout explaining that React or jsdom doesn't seem to correctly trigger an update on reset clicks. It would be extra nice if you can find an issue open in either project that covers that case.

pcberg:
@burnumd I did add a comment with Rishi's help. I looked for open issues and the best on "reset" having problems I found is this: https://github.com/algolia/instantsearch/issues/4160#issuecomment-539402090 but it doesn't fit because we already at on jest 25+

I'll go ahead and merge, let me know if you find anything else. Thank you!

PR #346, opened 2023-01-18
fix: 2.0.0 Fix Col tests by renaming width prop that conflicted with Rivetize

athill:


burnumd (changes requested):
Rather than removing the ability to rivetize this component, can we just change the name of the property in this component from `width` to `columnWidth`?

athill:
@burnumd Fixed. Thanks.

burnumd (approved):

PR #347, opened 2023-02-02
feat: 2.0.0 List component

athill:


src/components/List/List.jsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <ListTag id={id} className={classes} {...attrs}>
+      {listItems}
+    </ListTag>
+  );
+};
+
+List.displayName = "List";
+List.propTypes = {
+  /** A unique identifier for the alert */
```
burnumd, line 40:
```suggestion
  /** A unique identifier for the list */
```

burnumd (dismissed):
Before we merge this or any similar PR, I'd like to discuss the [alternatives](https://beta.reactjs.org/reference/react/Children#alternatives) to using the Children API since the React team recommends against child manipulation. We already have examples of things like the Grid and its associated Row/Col components where the parent component expects specific types for the children.

pcberg:
> Before we merge this or any similar PR, I'd like to discuss the [alternatives](https://beta.reactjs.org/reference/react/Children#alternatives) to using the Children API since the React team recommends against child manipulation. We already have examples of things like the Grid and its associated Row/Col components where the parent component expects specific types for the children.

I don't have enough experience for an insightful comment. But FWIW: I'm in support of following the recommendation.

src/components/List/DescriptionList.jsx
```diff
@@ -0,0 +1,48 @@
...
+  children
+    ? React.Children.map(children, (child, index) => {
+        if (
+          typeof child === "string" ||
+          typeof child === "number" ||
+          !["dd", "dt"].includes(child.type)
+        ) {
+          return index % 2 === 0 ? <dt>{child}</dt> : <dd>{child}</dd>;
+        }
+        return child;
```
ri-pandey, line 44: In regards to not using the Children API:

1. In case where client sends `<dt>` or `<dd>` elements inside the component, you can just return `children` as it is.
2. If they want to send an array, you could do something similar to [this pattern](https://beta.reactjs.org/reference/react/Children#accepting-an-array-of-objects-as-a-prop).

basham: I don't think that inspecting or modifying the children is needed in the DescriptionList or List components. Especially for `dl`, there are lots of valid arrangements that makes managing it likely more difficult than it is worth. For example, you can have multiple `dd` following a `dt`, not just zero or one. `dt` and `dd` elements can be grouped with `div`, meaning a `div` would be the direct child of the `dl`.

If the user wants to use an array or object to generate the children, that's fine. But it is a user concern, not a library concern. For example:

```jsx
<!-- Not this -->
<List>{["Item 1", "Item 2", "Item 3"]}</List>

<!-- Recommended approach -->
<List>
  {["Item 1", "Item 2", "Item 3"].map((item) => (<li>{item}</li>)}
</List>
```

basham (approved):

PR #351, opened 2023-02-23
326: Loading Indicator

pcberg:
* Question: Do we like the component name "LoadingIndicator"?
* I added the Button loading indicator sample to both Button.md and LoadingIndicator.md. The other changes in Button.md is cleanup, prettify seemed to have added some end of line weirdness. Hopefully this sticks

### DIFF
--- a/src/components/Button/Button.jsx
+++ b/src/components/Button/Button.jsx
@@ -6,6 +6,7 @@ import classNames from "classnames";
 import * as PropTypes from "prop-types";
 import * as React from "react";
 import * as Rivet from "../util/Rivet";
+import { LoadingIndicator } from "../LoadingIndicator/index.jsx";
 
 const buttonClass = "rvt-button";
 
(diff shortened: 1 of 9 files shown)


src/components/Button/Button.jsx
```diff
@@ -45,15 +51,20 @@ const Button = ({
...
       className
     )}
     onClick={onClick}
     ref={innerRef}
+    disabled={loading}
+    aria-busy={loading}
     {...attrs}
   >
-    {children}
+    <span className="rvt-button__content" role="buttonContent">
```
ri-pandey, line 64: You could consider wrapping the button text in this span only if the loading prop is true, to mimic Rivet more closely.

pcberg: Good idea, thanks Rishi! I improved it.

ri-pandey (approved):

pcberg:
I merged this, as always I'm happy to revisit should anybody notice something after the fact. I realize my turn around on this PR was fast. Thanks!

PR #356, opened 2023-05-11
353 - Added wrapper for render fallback on error of rivetize comps

johglove:
Work to add Error Boundary for rivetize components.  The prop type validations however do not trigger the react library to call `getDerivedStateFromError` so while it would work for errors which trigger a JavaScript failure (like NPE) it does not work for component prop violations (missing required prop, invalid type).  

PR opened for review purposes until direction is decided regarding this issue.

### DIFF
--- a/src/components/util/ErrorBoundary.jsx
+++ b/src/components/util/ErrorBoundary.jsx
@@ -1,0 +1,29 @@
+import React from "react";
+export default class ErrorBoundary extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      hasError: false,
+    };
+  }
+
+  static getDerivedStateFromError(error) {
+    console.log("Error set");
+    return { hasError: true };
+  }
+
+  componentDidCatch(error, info) {
+    console.log(error);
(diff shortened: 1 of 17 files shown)


src/components/util/ErrorBoundary.jsx
```diff
@@ -0,0 +1,29 @@
...
+    };
+  }
+
+  static getDerivedStateFromError(error) {
+    console.log("Error set");
+    return { hasError: true };
+  }
+
+  componentDidCatch(error, info) {
+    console.log(error);
```
basham, line 16: Should either of these use [`console.error()`](https://developer.mozilla.org/en-US/docs/Web/API/console/error) rather than `console.log()`?

basham (commented):
Also: I'm noticing a number of warnings when I run `npm run test`. I don't know if that should be fixed in this work or another PR.

burnumd:
Closing because after discussion, we found the solution would be better served by pointing end users of the library to Error Boundary documentation for their components which have more complex business logic.

PR #361, opened 2023-07-13
feat: update build process

burnumd:
* Use codecov CircleCI orb instead of Node dependency
* Export additional modules for more complete build
* Update Rivet-core and Vite


pcberg (approved):

johglove (approved):
Once branches are mostly cleaned up might want to look at removing package-lock and adding it to the git ignore.

Issue #375, opened 2023-07-24
2.x Input Group Component


burnumd:
There should be an [input group](https://rivet.iu.edu/components/input-group/) component

PR #387, opened 2023-08-15
feat: modify inputs for group functionality (#375)

johglove:


burnumd (changes requested):
As a reminder, commits must follow the [Angular Convention](https://www.conventionalcommits.org/en/v1.0.0-beta.4/#specification) for the release process to work

johglove:
What are the conventions not being followed?

burnumd:
> What are the conventions not being followed?

This PR, as an example, probably needs to be a `feat: modify inputs for group functionality (#375)`.
If it doesn't have a commit type, it will not trigger a release and the commit will not be included in the release notes.

src/components/Input/common.jsx
```diff
@@ -46,22 +54,29 @@ export const renderInput = (
...
     ...attrs
   }
 ) => {
   const noteId = `${id}_note`;
+  const groupped = appendment ||  prependment;
+
   const inputProps = {
     id,
-    className: inputClassName(elementName, variant),
+    className: inputClassName(elementName, variant, groupped),
```
burnumd, line 72:
```suggestion
    className: inputClassName(elementName, variant, grouped),
```

src/components/Input/common.jsx
```diff
@@ -73,7 +88,15 @@ export const renderInput = (
       >
         {label}
       </label>
-      {React.createElement(elementName, inputProps)}
+      {!groupped && inputElement}
```
burnumd, line 91:
```suggestion
      {!grouped && inputElement}
```

burnumd (changes requested):

burnumd (approved):

Issue #390, opened 2023-08-25
Change Export to Individual Components


johglove:
Current setup of `main.js` exports multiple components as groups rather than individual components.  For example Header in a collection of the components exported from `Header/index.js` rather than the Header component directly.  This prevents users from being able in import and user the component directly.  Instead the collection imported has to be deconstructed to get individual components. 

Example of current usage:
```
import { Header as RRHeader } from 'rivet-react'
const  { Header, HeaderAvatar, HeaderMenu } = RRHeader
```

Expected usage:
```
import { Header, HeaderAvatar, HeaderMenu } from 'rivet-react'
```

PR #397, opened 2023-08-30
feat: Export Individual Components (#390)

johglove:


lib/main.js
```diff
@@ -2,38 +2,43 @@
 Copyright (C) 2018 The Trustees of Indiana University
 SPDX-License-Identifier: BSD-3-Clause
 */
-import * as Alert from '../src/components/Alert/index';
+import * as RRAlert from '../src/components/Alert/index';
```
basham, line 5: I suspect you don't need to do this `import` then `export` step. You may be able to do it in one line. Have you tried this? This approach would make this file easier to maintain.

```js
export * from '../src/components/Alert/index';
```

basham (changes requested):

lib/main.js
johglove: This appears to work.  It does have the drawback is a loss of purpose control on exports (components will be automatically instead of having a speedbump here).  Can even shorten it to `export * from '../src/components/index';` but would lose even more purpose control.

basham (approved):

Issue #392, opened 2023-08-25
Allow for Configuration of Logout in HeaderAvatar


johglove:
Currently the HeaderAvatar component only allows for the setting of a url for the logout action/button.  With the increase in single page and on page actions the component needs to support configuring JavaScript/events for the logout.

PR #399, opened 2023-09-07
feat: Header Avatar Log out Enhancement (#392) 

johglove:


src/components/Header/HeaderAvatar.jsx
```diff
@@ -42,6 +62,8 @@ HeaderAvatar.propTypes = {
   shortName: PropTypes.string.isRequired,
   /** The URL for the logout action */
   logoutURL: PropTypes.string,
+  /** The onClick for the logout action*/
```
burnumd, line 65: The docs should probably indicate that `logoutClick` and `logoutUrl` are mutually exclusive:

```suggestion
  /** The onClick for the logout action. This value will be overriden by logoutUrl if it is provided  */
```

burnumd (changes requested):

burnumd (approved):

Issue #402, opened 2023-09-21
Setup Header Components with Processing


johglove:
Current Header components have automatic processing and features that enforce specific functionality and features.  For more customizable headers need corresponding components that allow for manual setup and assume user responsibility for correct syntax.

PR #403, opened 2023-09-21
feat: Implemented Base Header Components (#402)

johglove:
The base versions of the header components remove processing and a good portion of the enforced standards that the normal components have.  The base versions are meant give users pieces they can then put together to make their own header with fewer restrictions.  It does put more responsibility on the user to ensure components are semantically correct as a trade off.


src/components/Header/HeaderAvatar.jsx
```diff
@@ -20,7 +20,7 @@ const HeaderAvatar = ({ username, shortName, logoutURL }) => {
         </span>
       </div>
       <div
-        className="rvt-ts-14 rvt-m-left-xs rvt-p-right-xs rvt-m-right-xs rvt-border-right"
+        className={`rvt-ts-14 rvt-m-left-xs rvt-p-right-xs rvt-m-right-xs${logoutURL ? " rvt-border-right" : ""}`}
```
burnumd, line 23: This seems like a potential use case for the classnames library.

PR #404, opened 2023-09-21
feat: Disclosure Component

pcberg:


src/components/PageContent/Disclosure/Disclosure.jsx
```diff
@@ -0,0 +1,115 @@
...
+    <div
+      id={id}
+      className={classNames(disclosureClass, className)}
+      ref={disclosureWrapDiv}
+      {...attrs}
+    >
+      <button
+        className="rvt-disclosure__toggle"
+        onClick={toggleDisclosure}
+        aria-expanded="false"
```
johglove, line 87: Believe this needs to reflect the open state of the component.

johglove (changes requested):
Currently the chevron icon added to the title does not reflect the open state of the disclosure.

src/components/PageContent/Disclosure/Disclosure.jsx
pcberg: @johglove thanks for catching that, I fixed it. Let me know what you think!

burnumd (approved):

johglove (approved):

Issue #374, opened 2023-07-24
2.x Hero Component


burnumd:
There should be a [Hero](https://rivet.iu.edu/components/hero/) component.

PR #412, opened 2023-11-28
feat: Implement Hero Component (#374)

johglove:


src/components/PageContent/Hero/Hero.jsx
```diff
@@ -0,0 +1,111 @@
...
+Hero.displayName = "Hero";
+Hero.propTypes = {
+  /** Available actions for the element*/
+  actions: PropTypes.element,
+  /** A unique identifier for the element */
+  id: PropTypes.string,
+  /** Image or media element to display  */
+  media: PropTypes.element,
+  /** Text to accompany a provided media */
+  mediaCaption: PropTypes.string,
```
burnumd, line 102: What is the difference between the `mediaCaption` and the `children` of this component?

johglove: Media caption is a text line that can be added to the section containing the media part (guessing usually and image) while the children is the text content of the whole component.  Media caption is also reliant on a media being provided and won't display by itself.

burnumd: Would you mind adding that clarification to the docs on line 101?

burnumd (approved):

Issue #372, opened 2023-07-24
2.x Card Component


burnumd:
There should be a [card](https://rivet.iu.edu/components/card/) component.

PR #420, opened 2024-01-24
feat: Implemented Card component (#372)

johglove:


src/components/PageContent/Card/Card.test.jsx
```diff
@@ -0,0 +1,255 @@
...
+  expect(elemet).toHaveClass("rvt-card__eyebrow");
+  expect(elemet.innerHTML).toBe(eyebrow);
+}
+
+const checkRenderMeta = () => {
+  const elemet = screen.getByTestId(testIds.meta);
+  expect(elemet).toBeVisible();
+  expect(elemet).toHaveClass("rvt-card__meta");
+  expect(elemet.innerHTML).toBe(metaStr);
+}
```
burnumd, line 255:
```suggestion
}

```

src/components/PageContent/Card/Card.jsx
```diff
@@ -0,0 +1,119 @@
...
+  horizontal = false,
+  image,
+  meta,
+  raised = false,
+  testMode = false,
+  title,
+  titleUrl,
+  ...attrs
+}) => {
+  const classNameArr = [
```
burnumd, line 26: Should we do any validation that if `clickable` is `true` that `title` and `titleUrl` must not be undefined?

burnumd (changes requested):

src/components/PageContent/Card/Card.jsx
johglove: The title is already required from the prop types and enforcing that with code hasn't been a practice (most checks are to prevent javascript errors).  This would be similar with the titleUrl, if the url isn't set then the clickable doesn't class doesnt do anything (card acts like a normal without link).

burnumd (approved):

Issue #385, opened 2023-07-24
2.x Tabs Component


burnumd:
There should be a [tabs](https://rivet.iu.edu/components/tabs/) component.

PR #421, opened 2024-01-26
feat: Implemented Tabs component (#385)

johglove:


src/components/Tabs/Tabs.jsx
```diff
@@ -0,0 +1,139 @@
...
+  id: PropTypes.string,
+  /** An unseen label for the menu to help with accessibility */
+  label: PropTypes.string.isRequired,
+  /** Index of initially opened tab */
+  initialTab: PropTypes.number,
+  /** [Developer] Adds data-testId attributes for component testing */
+  testMode: PropTypes.bool
+};
+
+Tabs.Panel = TabPanel;
```
pcberg, line 91: Thoughts on Tab.Panel?

pcberg (changes requested):

src/components/Tabs/Tabs.jsx
johglove: The Tabs part is the parent component which is the container of multiple tabs so the name Tab wouldn't really work.

pcberg (approved):

Issue #382, opened 2023-07-24
2.x Step Indicator Component


burnumd:
There should be a [step indicator](https://rivet.iu.edu/components/step-indicator/) component.

PR #422, opened 2024-02-01
feat: Implemented Step Indicator (#382)

johglove:


src/components/PageContent/StepIndicator/StepIndicator.jsx
```diff
@@ -0,0 +1,49 @@
...
+import * as React from "react";
+import * as Rivet from "../../util/Rivet";
+import { TestUtils } from "../../util/TestUtils";
+import Step from "./Step"
+
+const testIds = TestUtils.StepIndicator
+/**
+ * Show the user's position in and progress through a multi-step process
+ */
+const StepIndicator = ({
```
pcberg, line 16: Thoughts on calling this Steps instead of StepIndicator? Or I suppose StepsIndicator but personally I like Steps more...

The more I look at it, it also seems to me that the indicator is the nested element... anyway, my vote is Steps/Step. :)

src/components/PageContent/StepIndicator/Step.jsx
```diff
@@ -0,0 +1,71 @@
...
+}) => {
+  const classNameArr = [
+    "rvt-steps__item",
+    className
+  ]
+  const indicatorClassNameArr = [
+    "rvt-steps__indicator",
+    variant === "success" ? "rvt-steps__indicator--success" : "",
+    variant === "warning" ? "rvt-steps__indicator--warning" : "",
+    variant === "danger" ? "rvt-steps__indicator--danger" : ""
```
pcberg, line 32: You could replace this with
`variant ? `rvt-switch--${variant}` : "",`

pcberg (changes requested):

src/components/PageContent/StepIndicator/StepIndicator.jsx
johglove: I originally debated on the naming too but StepIndicator keeps the primary component inline and connected to the rivet component. While the Steps matches the css of the container component the two components are not really for independent use.  The name StepIndicator does describe the combined functionality and even if the Step components are switch out for custom the container is still the start of the Indicator.

pcberg (approved):

Issue #367, opened 2023-07-24
2.x Accordion component


burnumd:
There should be a Rivet 2 [accordion](https://rivet.iu.edu/components/accordion/) component.

PR #423, opened 2024-02-05
feat: Implemented Accordion (#367)

johglove:


src/components/Accordion/Accordion.md
```diff
@@ -0,0 +1,56 @@
...
+
+<!-- prettier-ignore-start -->
+```jsx
+<Accordion id="test1" initial="1" iconClosed={<Icon name="caret-down" />} iconOpened={<Icon name="toggle-close" />}>
+    <Accordion.Panel title="Section One">
+        <p>Some content text</p>
+    </Accordion.Panel>
+    <Accordion.Panel title="Section Two">
+        <p>Panel starts open as indicated by initial property.</p>
+        <p>Additionally the icons in the headers are set in the propertys rather than the default option
```
pcberg, line 33:
```suggestion
        <p>Additionally the icons in the headers are set in the propertys rather than the default option</p>
```

src/components/Accordion/Accordion.jsx
```diff
@@ -0,0 +1,115 @@
...
+const testIds = TestUtils.Accordion
+
+/**
+ * Group content into sections that can be opened and closed
+ */
+const Accordion = ({
+  children,
+  className,
+  iconClosed,
+  iconOpened,
```
pcberg, line 24: Can we default these two per the docs sample? Thoughts?

Seems like that covers most usual cases...

pcberg (changes requested):

pcberg (approved):

Issue #425, opened 2024-02-09
BaseHeaderMenu Toggle not Working in Rivet 18 Due to Double Click


johglove:
Need to add preventDefault and stopPropagation to toggle menu action to prevent double clicks

PR #426, opened 2024-02-09
feat: Added preventDefault/stopPropagation to toggle action (#425)

johglove:


pcberg (approved):
Thanks, we tested this in EBS-FS' alog project, that will help us!

PR #433, opened 2024-05-16
fix: Cleaned up prop type console error warnings

johglove:


src/components/Header/BaseHeaderMenu.jsx
```diff
@@ -187,7 +187,7 @@ BaseHeaderMenu.propTypes = {
   /** Indicates item is current page if link with url */
   current: PropTypes.bool,
   /** The label of the menu */
-  label: PropTypes.string.isRequired,
+  label: PropTypes.oneOfType([PropTypes.string, PropTypes.object]).isRequired,
```
burnumd, line 190: This seems redundant. Is it intended that it's a `PropTypes.node`?

src/components/PageContent/Billboard/Billboard.jsx
```diff
@@ -47,7 +47,7 @@ const Billboard = ({
 Billboard.displayName = "Billboard";
 Billboard.propTypes = {
   /** Optional image to display */
-  image: PropTypes.oneOfType(PropTypes.element),
+  image: PropTypes.oneOfType([PropTypes.element]),
```
burnumd, line 50: I'm not sure why these were `oneOfType` to begin with if we only accept one type.

burnumd (approved):

PR #435, opened 2024-05-16
test: Added additional test code coverage

johglove:
Adds coverage to all lines except the following:
BaseHeaderNavigationSecondary.jsx   |     100 |    96.15 |     100 |     100 | 21
HeaderNavigationSecondary.jsx       |     100 |    93.75 |     100 |     100 | 22
HeaderSearch.jsx                    |     100 |    92.85 |     100 |     100 | 17

The remaining lines should be covered as they reflect functionality and tests similar to the HeaderNavigation component but for some reason are not picked up as covered like it is.

childUtils.jsx                      |     100 |    94.44 |     100 |     100 | 26
Attempting to add coverage for this test causes the component to error if the li wrappers are removed from the anchor elements.


src/components/PageContent/Disclosure/Disclosure.jsx
```diff
@@ -56,8 +56,7 @@ const Disclosure = ({
       // If the user right clicks anywhere on the screen or they press an unhandled key do not close the menu
       return false;
     } else if (
-      targets(disclosureWrapDiv.current, event) &&
-      (!isKeyEvent(event) || isTabKeyPress(event))
+      targets(disclosureWrapDiv.current, event)
```
johglove, line 59: Removing check does not appear to effect functionality and report showed the && was not covered.  Appears that anything that would trigger targets as true would also have the right side of the && as true.

burnumd (approved):

Issue #436, opened 2024-05-23
Issues with the Header Menu


johglove:
There are a couple issues observed with the Header Menu components.  These were seen with use of the BaseHeaderMenu but may extend to the normal version as well.
1. If you have multiple menus and have one opened when another one is opened the first one remains open instead of closing as expected.
2. Clicking a link in the menu does not execute as expected unless stop propagation is called on the link (BaseHeaderMenu Only)

johglove:
Completed by https://github.com/indiana-university/rivet-react/pull/437

PR #437, opened 2024-05-23
fix: Corrected issues with Base Header Menu Component (#436)

johglove:


burnumd (approved):
Can you add a description linking to the appropriate issue on the PR?

PR #463, opened 2024-09-20
fix: 462 - Fix Header.Avatar display - closes #462

### LINKED ISSUE #462: Issues using Header.Avatar
Header.Avatar has a few issues:

1. When used in a header, it should be documented that it goes inside `BaseHeaderNavigation`.
2. The avatar component itself is in a div, but when included in header navigation, it is erroneously put inside the `ul` tag

It is likely that the Avatar information should be included as props to `BaseHeader` instead of being provided as a component since the positioning of it is highly specific. Either that, or detect when an `Header.Avatar` component is included as a child of `BaseHeaderNavigation` and place it outside the generated `ul` tag.

athill:
The `Header.Avatar` and `Header.Search` components need to be rendered inside the `BaseHeaderNavigation`'s `nav` component, but outside of the navigation `ul`, which was not happening previously.

This change allows the search and avatar components to be added as children to `BaseHeaderNavigation` and the various types of components are placed in their respective locations.

### DIFF
--- a/src/components/Alert/Alert.jsx
+++ b/src/components/Alert/Alert.jsx
@@ -15,7 +15,7 @@ import { TestUtils } from "../util/TestUtils.js";
 const Alert = ({
   title,
   onDismiss,
-  variant,
+  variant = "info",
   isOpen = true,
   id = Rivet.shortuid(),
   className,
@@ -88,9 +88,9 @@ Alert.propTypes = {
   /** [Developer] Adds data-testId attributes for component testing */
   testMode: PropTypes.bool,
   /** An extremely brief title for the alert */
-  title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
+  title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
   /** The variant type which determines how the alert is styled */
-  variant: PropTypes.oneOf(["info", "success", "warning", "danger"]).isRequired,
+  variant: PropTypes.oneOf(["info", "success", "warning", "danger"]),
 };
 
 export default Rivet.rivetize(Alert);
--- a/src/components/File/File.test.jsx
+++ b/src/components/File/File.test.jsx
@@ -88,7 +88,9 @@ describe("<File />", () => {
       expect(description.innerHTML).toBe("2 files selected");
     });
 
-    it("should allow resetting if embedded in a form", async () => {
+    // This test is flaky and unreliable. We should manually verify the expected behavior when
+    // making changes or updating dependencies
+    it.skip("should allow resetting if embedded in a form", async () => {
       const fileName = "foo.txt";
       let fileObject = new File(["foo"], fileName, { type: "text/plain" });
       const resetId = "reset-id";
--- a/src/components/File/File.test.jsx
+++ b/src/components/File/File.test.jsx
@@ -91,13 +91,14 @@ describe("<File />", () => {
     it("should allow resetting if embedded in a form", async () => {
       const fileName = "foo.txt";
       let fileObject = new File(["foo"], fileName, { type: "text/plain" });
+      const resetId = "reset-id";
       render(
         <form>
           <MyFile
             id={TestUtils.File.testId}
             data-testid={TestUtils.File.testId}
           />
-          <input type="reset" data-testid="reset-id" value="Reset" />
+          <input type="reset" data-testid={resetId} value="Reset" />
         </form>
       );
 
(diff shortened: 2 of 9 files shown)


src/components/Alert/Alert.jsx
```diff
@@ -88,9 +88,9 @@ Alert.propTypes = {
   /** [Developer] Adds data-testId attributes for component testing */
   testMode: PropTypes.bool,
   /** An extremely brief title for the alert */
-  title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
```
athill, line 91: There is a test that `title` is not required, but it was still required. This fixes that.

src/components/Header/BaseHeader.md
```diff
@@ -62,9 +62,9 @@ const secondaryNavigation = <BaseHeaderNavigationSecondary title="Component Libr
                     <BaseHeaderMenuItem itemUrl="#" subMenu>Sub item three</BaseHeaderMenuItem>
                 </BaseHeaderMenu>
             </BaseHeaderMenuItem>
+            <Header.Search action={"/mySearchURL"} method={"post"} />
+            <Header.Avatar username={"johndoe"} shortName={"jd"} logoutURL={"/logout"} />
     </BaseHeaderNavigation>
-    <Header.Search action={"/mySearchURL"} method={"post"} />
-    <Header.Avatar username={"johndoe"} shortName={"jd"} logoutURL={"/logout"} />
```
athill, line 67: `Header.Search` and `Header.Avatar` need to be children of `BaseHeaderNavigation`

src/components/Header/BaseHeader.test.jsx
```diff
@@ -62,24 +62,29 @@ describe("<BaseHeader />", () => {
...
-            </BaseHeaderMenu>
+            <BaseHeaderMenuItem>
+              <BaseHeaderMenu label="Nav item three">
+                <BaseHeaderMenuItem href={"#"}>Sub item one</BaseHeaderMenuItem>
+                <BaseHeaderMenuItem href={"#"}>Sub item two</BaseHeaderMenuItem>
+                <BaseHeaderMenuItem href={"#"}>
+                  Sub item three
+                </BaseHeaderMenuItem>
+              </BaseHeaderMenu>
+            </BaseHeaderMenuItem>
```
athill, line 87: `BaseHeaderMenu` needs to be wrapped in a `BaseHeaderMenuItem` for its output to match the `rivet` examples. Having it as a child of `BaseHeaderNavigation` is now disallowed via its prop types

src/components/Header/BaseHeaderNavigation.jsx
```diff
@@ -84,9 +99,8 @@ const BaseHeaderNavigation = ({ children, testMode = false, ...attrs }) => {
         hidden={!isNavMenuOpen}
         {...(testMode && { "data-testid": TestUtils.Header.headerNavTestId })}
       >
-        <ul className={"rvt-header-menu__list"}>
-          {children}
-        </ul>
+        <ul className={"rvt-header-menu__list"}>{listItems}</ul>
```
burnumd, line 102:
```suggestion
        <ul className="rvt-header-menu__list">{listItems}</ul>
```

src/components/Header/BaseHeaderNavigation.test.jsx
```diff
@@ -22,21 +24,15 @@ describe("<BaseHeaderNavigation />", () => {
...
-          </ul>
+          <BaseHeaderMenuItem itemUrl="#">Nav item one</BaseHeaderMenuItem>
+          <BaseHeaderMenuItem itemUrl="#">Nav item two</BaseHeaderMenuItem>
+          <BaseHeaderMenuItem>
+            <BaseHeaderMenu label="Nav item three">
+              <a href={"#"}>Sub item one</a>
+              <a href={"#"}>Sub item two</a>
+              <a href={"#"}>Sub item three</a>
+            </BaseHeaderMenu>
+          </BaseHeaderMenuItem>
```
athill, line 35: `BaseHeaderNavigation` now restricts the component types it allows as its children. This was also incorrect HTML, as `BaseHeaderNavigation` generates a `ul` tag, so this basically had nested `ul` tags.

src/components/Header/BaseHeaderNavigation.jsx
```diff
@@ -65,14 +67,27 @@ const BaseHeaderNavigation = ({ children, testMode = false, ...attrs }) => {
     }
   };
 
+  // Avatar and Search components need to be rendered inside the nav tag, but outside of the list
+  const listItems = [];
+  const otherHeaderMenuItems = [];
+  React.Children.forEach(children, (child) => {
+    if ([Header.Avatar, Header.Search].includes(child.type)) {
```
burnumd, line 74: I think it may make more sense to reverse the sense of this test. That is, if the child is a `BaseHeaderMenuItem` add it to the `listItems` and otherwise it goes outside the list. Since we know `BaseHeaderMenuItem` is wrapped in a `li`, it can go in the list and any additional content is outside the list content.

burnumd (changes requested):

src/components/Header/BaseHeaderNavigation.jsx
athill: Works for me. Done. Thanks.

burnumd (approved):

src/components/File/File.test.jsx
```diff
@@ -88,7 +88,9 @@ describe("<File />", () => {
       expect(description.innerHTML).toBe("2 files selected");
     });
 
-    it("should allow resetting if embedded in a form", async () => {
+    // This test is flaky and unreliable. We should manually verify the expected behavior when
+    // making changes or updating dependencies
+    it.skip("should allow resetting if embedded in a form", async () => {
```
burnumd, line 93: @pcberg @johglove, I wanted to double check that y'all are ok with this change. It narrowly impacts coverage, but the test failing only occasionally impacts development and DX and this branch can be manually tested during dependency updates.

pcberg: No objections from me.

Issue #467, opened 2024-10-11
Skip to content link always takes user to `/#main-content` instead of the page they were on


burnumd:
The [skip to content link](https://github.com/indiana-university/rivet-react/blob/d9be3052e05ad905f2b3bf422882351dc04dc369/src/components/Header/BaseHeader.jsx#L41) is hard coded to link to the main content on the index page. It should instead link to the main content anchor on URL the user was at. It looks like this was fixed in the Rivet-React 1.x code [here](https://github.com/indiana-university/rivet-react/pull/194/files)

PR #468, opened 2024-10-17
feat: #467 make content href customizable

burnumd:
Make the content href a required property on `Header` and `BaseHeader` so it is enforced that users set it consciously.

### DIFF
--- a/src/components/Header/BaseHeader.jsx
+++ b/src/components/Header/BaseHeader.jsx
@@ -13,6 +13,7 @@ import { TestUtils } from "../util/TestUtils";
 const BaseHeader = ({
   children,
   className,
+  contentHref,
   title,
   subtitle,
   headerWidth = "fluid",
@@ -38,7 +39,7 @@ const BaseHeader = ({
     <header {...attrs} className={classNames("rvt-header-wrapper", className)}>
       <a
         className="rvt-header-wrapper__skip-link"
-        href={`${document.url}#main-content`}
+        href={contentHref}
         {...(testMode && { "data-testid": TestUtils.Header.skipLinkTestId })}
       >
         Skip to main content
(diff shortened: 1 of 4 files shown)


src/components/Header/BaseHeader.jsx
```diff
@@ -89,6 +90,8 @@ BaseHeader.displayName = "BaseHeader";
 BaseHeader.propTypes = {
   /** The application name or title that appears in the header */
   title: PropTypes.string.isRequired,
+  /** The href to use for the skip to content link */
+  contentHref: PropTypes.string.isRequired,
```
basham, line 94: `contentHref` shouldn't be required. It should default to `#main-content`. That will make it more backwards compatible.

basham (changes requested):

src/components/Header/BaseHeader.jsx
basham: Although I do get your point that by forcing users to explicitly set it, it makes them more aware of it and perhaps more likely to forget to set `<main id>`.

I'd welcome other thoughts on this.

@johglove, @pcberg?

burnumd: I'm less concerned about backwards compatibility in beta, especially because we don't enforce the ID `#main-content` on the `main` tag and don't document anywhere that it's what the header expects. This change will only affect users who are upgrading within the beta and I think making it required and documenting it for new users provides more value.

basham (approved):

src/components/Header/BaseHeader.jsx
pcberg: Looks like in rivet-react v1 it was defaulted https://github.com/indiana-university/rivet-react/blob/96b2a2d343932f4dc2870c87fdea77cb6f06daa7/src/components/Header/Header.tsx#L33 and we seemed to have noticed for the projects where we use it on our team.

Having said that, if it was me I would not have noticed so an isRequired would have made me aware. I'm in favor of that approach. But I can be convinced either way.

pcberg (approved):

PR #472, opened 2024-10-31
feat: Add onClick behavior to BaseHeader and BaseHeaderMenuItem components - closes #471

### LINKED ISSUE #471: Add onClick behavior to `BaseHeader` and `BaseHeaderMenuItem` components
The `BaseHeader` and `BaseHeaderMenuItem` components contain links that are rendered as anchor tags. This will add optional onClick propertiesto the anchor tags so that default behavior can be overwritten, for example, to use `react-router`'s `navitgate()`

athill:
Adding onClick handlers to a couple components so that we can override the default behavior in their nested anchor tags and use, for example, `react-router`'s `navigate()` function.

### DIFF
--- a/src/components/Header/BaseHeader.jsx
+++ b/src/components/Header/BaseHeader.jsx
@@ -36,6 +36,12 @@ const BaseHeader = ({
       </div>
     </>
   );
+  const homeUrlOnClick = onClick
+    ? (e) => {
+        e.preventDefault();
+        onClick(e);
+      }
+    : null;
   return (
     <header {...attrs} className={classNames("rvt-header-wrapper", className)}>
       <a
@@ -58,7 +64,7 @@ const BaseHeader = ({
                 <a
                   className="rvt-lockup"
                   href={homeUrl}
-                  onClick={onClick}
+                  onClick={homeUrlOnClick}
                   {...(testMode && {
                     "data-testid": TestUtils.Header.anchorTestId,
                   })}
@@ -94,7 +100,7 @@ BaseHeader.propTypes = {
   title: PropTypes.string.isRequired,
   /** The href to use for the skip to content link */
   contentHref: PropTypes.string.isRequired,
-  /** Optional onClick handler that will be applied to the homeUrl anchor tag */
+  /** Optional onClick handler that will be applied to the homeUrl anchor tag, it will prevent the anchor tag's default behavior */
   onClick: PropTypes.func,
   /** Optional subTitle that will be displayed below the title */
   subtitle: PropTypes.string,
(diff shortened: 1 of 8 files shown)


src/components/Header/BaseHeaderMenuItem.jsx
```diff
@@ -47,10 +59,12 @@ BaseHeaderMenuItem.propTypes = {
   current: PropTypes.bool,
   /** The navigation url for the item */
   itemUrl: PropTypes.string,
+  /** Optional onClick handler for itemUrl anchor tag */
+  onClick: PropTypes.func,
   /** Menu item is part of a sub menu */
   subMenu: PropTypes.bool,
   /** [Developer] Adds data-testId attributes for component testing */
-  testMode: PropTypes.bool
+  testMode: PropTypes.bool,
```
athill, line 67: Many of the changes in this file are due to Prettier. The only substantive change is adding the `onClick` property.

src/components/Header/BaseHeaderMenuItem.test.jsx
```diff
@@ -14,17 +14,15 @@ import userEvent from "@testing-library/user-event";
 const user = userEvent.setup();
 
 describe("<BaseHeaderMenuItem/>", () => {
-  describe("Rendering and styling",  () => {
```
athill, line 17: Many of the changes in this file are due to Prettier. The only substantive change is adding the onClick property test.

johglove (approved):

src/components/Header/BaseHeader.jsx
```diff
@@ -57,6 +58,7 @@ const BaseHeader = ({
                 <a
                   className="rvt-lockup"
                   href={homeUrl}
+                  onClick={onClick}
```
burnumd, line 61: What if we handled the `preventDefault` behavior *only when* `onClick` is specified so end users don't have to do so?

burnumd (changes requested):

src/components/Header/BaseHeader.jsx
athill: Done

src/components/Header/BaseHeader.jsx
```diff
@@ -57,6 +64,7 @@ const BaseHeader = ({
                 <a
                   className="rvt-lockup"
                   href={homeUrl}
+                  onClick={homeUrlOnClick}
```
burnumd, line 67: We only want to prevent default if the `onClick` is set.

```suggestion
                  onClick={onClick && homeUrlOnClick}
```

burnumd (changes requested):

src/components/Header/BaseHeader.jsx
athill: `homeUrlOnClick` will be `null` if `onClick` is falsy. See the code where it is set up.

burnumd: Ah, I see that now, thank you.

burnumd (approved):

PR #483, opened 2025-01-06
feat: #398 use rivet-icons v3

### LINKED ISSUE #398: Use Rivet Icons v3
The upcoming [Rivet Icons v3](https://github.com/indiana-university/rivet-icons/pull/105) will make it much easier to make custom icon sets and include custom icons. Once that is released (likely September or October 2023), explore if the [`<Icon>` React component](https://github.com/indiana-university/rivet-react/blob/v2.0.0-alpha.2/src/components/util/RivetIcons.jsx) can be replaced by the `<rvt-icon>` custom element.

burnumd:
install `rivet-icons@3.0.0` and use it in every place where we were previously using custom SVGs (where possible).

Closes #398

### DIFF
--- a/src/components/Accordion/Accordion.jsx
+++ b/src/components/Accordion/Accordion.jsx
@@ -20,8 +20,6 @@ const testIds = TestUtils.Accordion;
 const Accordion = ({
   children,
   className,
-  iconClosed,
-  iconOpened,
   id,
   initial,
   testMode = false,
@@ -31,7 +29,7 @@ const Accordion = ({
 
   const [accordianId] = useState(id);
   const [panelsOpen, setPanelsOpen] = useState(
-    getInitialOpen(initial, accordianId, numPanels)
+    getInitialOpen(initial, accordianId, numPanels),
   );
 
   const accordianPanels = !children
@@ -59,8 +57,6 @@ const Accordion = ({
         return {
           onClick,
           controlId,
-          iconClosed,
-          iconOpened,
           isOpen,
           label: title,
           panel,
@@ -90,10 +86,6 @@ const Accordion = ({
 
 Accordion.displayName = "Accordion";
 Accordion.propTypes = {
-  /** Icon/Indicator for a closed panel */
-  iconClosed: PropTypes.element,
-  /** Icon/Indicator for an open panel */
-  iconOpened: PropTypes.element,
   /** A unique identifier for the accordian */
   id: PropTypes.string.isRequired,
   /** Index of initially opened tab */
--- a/src/components/Accordion/AccordionPanelHeader.jsx
+++ b/src/components/Accordion/AccordionPanelHeader.jsx
@@ -43,7 +43,7 @@ const AccordionPanelHeader = (props) => {
       >
         <span className="rvt-accordion__toggle-text">{label}</span>
         <div className="rvt-accordion__toggle-icon">
-          {isOpen ? <rvt-icon name="minus" /> : <rvt-icon name="plus" />}
+          <rvt-icon name={isOpen ? "minus" : "plus"} />
         </div>
       </button>
     </h3>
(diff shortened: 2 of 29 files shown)


src/components/util/RivetIcons.jsx
```diff
@@ -38,33 +24,6 @@ const IconCore = ({ children, ...attrs }) => {
...
-const icoFile = (attrs) => (
-  <IconCore {...attrs}>
-    <path
-      fill="currentColor"
-      d="M10.41,1H3.5A1.3,1.3,0,0,0,2.2,2.3V13.7A1.3,1.3,0,0,0,3.5,15h9a1.3,1.3,0,0,0,1.3-1.3V4.39ZM11.8,5.21V6H9.25V3h.34ZM4.2,13V3h3V6.75A1.25,1.25,0,0,0,8.5,8h3.3v5Z"
-    />
-  </IconCore>
-);
-
 const icoLogo = (attrs) => (
```
basham, line 27: I'm unsure if this would work well in this case, but you could try using Rivet Icon's [`registerIcon()` utility](https://github.com/indiana-university/rivet-icons?tab=readme-ov-file#registericon-function) to add custom icons. That would allow you to do something like `<rvt-icon name="trident-header" />`.

basham (changes requested):
I did not run this branch locally. I'm just skimming through the code and providing some suggestions.

basham (approved):

src/components/Accordion/AccordionPanelHeader.jsx
```diff
@@ -16,8 +17,6 @@ const AccordionPanelHeader = (props) => {
     children,
     className,
     controlId,
-    iconOpened = <Icon name={IconNames.ACCORDIAN_IND} />,
```
johglove, line 19: Is there a reason to remove developers ability to set a custom icon?

src/components/Accordion/Accordion.jsx
```diff
@@ -20,8 +20,6 @@ const testIds = TestUtils.Accordion;
 const Accordion = ({
   children,
   className,
-  iconClosed,
```
johglove, line 23: Is there a reason to remove developers ability to set a custom icon?

src/components/Accordion/AccordionPanelHeader.jsx
burnumd: Part of the guidance on rivet-icons is to use them as they are ant not to make modifications to size/shape/etc. The customization option should probably not have been included in the first place but if there's a reason to deviate from the vanilla Rivet icons, let me know.

src/components/SeriesNav/SeriesNav.jsx
```diff
@@ -95,10 +91,13 @@ const SeriesControl = ({
         </span>
       </div>
       <span className="rvt-seriesnav__icon">
-        <Icon name={previous ? IconNames.ARROW_LEFT :  IconNames.ARROW_RIGHT} />
+        <rvt-icon
+          name={previous ? "arrow-left" : "arrow-right"}
+          {...(testMode && { "data-testid": TestUtils.RivetIcons.testId })}
```
johglove, line 96: Rather than using the RivetIcons component's version it would be better to define an unique id (ex. TestUtils.SeriesNav.controlIcon)

src/components/Accordion/Accordion.jsx
basham: Maybe @levimcg may want to chime in here. But I don't see a reason for this level of customization. The "plus" and "minus" icons are what [Rivet uses for accordions](https://rivet.uits.iu.edu/components/accordion/). I don't know of a case in which these icons should be anything else.

johglove (approved):

src/components/Accordion/Accordion.jsx
levimcg: @johglove @basham Yeah, we wouldn't ever really use a different icon here. We should always use the same icons for the accordion interaction to be consistent.

Issue #475, opened 2024-12-05
HeaderNavigation.jsx causes react warning/error:


rileymg:
The rivet-react component HeaderNavigation causes the following warning (which should prob. be an error as it can cause issues):

Warning: Each child in a list should have a unique "key" prop. See https://reactjs.org/link/warning-keys for more information. Error Component Stack
    at nav (<anonymous>)
    at div (<anonymous>)
...

We found the error disappeared when we removed the HeaderNavigation reference  from our app component AppHeader.jsx: https://github.iu.edu/iu-uits-es/ess-sis-stellic/blob/develop/web/src/main/js/components/AppHeader.jsx
and we are using     "rivet-react": "2.0.0-beta.15"

burnumd:
@rileymg Can you post your usage of the `HeaderNavigation` component (e.g., the code that's throwing the warning)?

burnumd:
Oh, I see it's in your link, I thought when you said you removed it, it wasn't there. I'm investigating the issue.

PR #484, opened 2025-01-08
fix: #475 fix header navigation child key warnings

burnumd:
* Use memoized child dependency in HeaderNavigation to determine when children change and key them with a UUID.
* Fix missing contentHref and icons in markdown documentation files.

### DIFF
--- a/src/components/Pagination/Pagination.md
+++ b/src/components/Pagination/Pagination.md
@@ -84,7 +84,7 @@ View the [Rivet documentation for Pagination](https://rivet.iu.edu/components/pa
         ariaLabel="Go to previous page"
         className="rvt-m-lr-sm"
     >
-        <Icon className="rvt-m-right-xs" name="previous" />
+        <rvt-icon className="rvt-m-right-xs" name="previous" />
         Back
     </PaginationItem>
     <PaginationItem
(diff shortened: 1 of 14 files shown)


src/components/Pagination/Pagination.md
```diff
@@ -84,7 +84,7 @@ View the [Rivet documentation for Pagination](https://rivet.iu.edu/components/pa
         ariaLabel="Go to previous page"
         className="rvt-m-lr-sm"
     >
-        <Icon className="rvt-m-right-xs" name="previous" />
+        <rvt-icon className="rvt-m-right-xs" name="previous" />
```
basham, line 87: There is no "previous" icon. This should be `chevron-left`.

src/components/Pagination/Pagination.md
```diff
@@ -116,7 +116,7 @@ View the [Rivet documentation for Pagination](https://rivet.iu.edu/components/pa
         className="rvt-m-lr-sm"
     >
         Next
-        <Icon className="rvt-m-left-xs" name="next" />
+        <rvt-icon className="rvt-m-left-xs" name="next" />
```
basham, line 119: There is no "next" icon. This should be `chevron-right`.

basham (changes requested):

basham (approved):

Issue #451, opened 2024-08-27
Consolidate and document use cases for Header components


burnumd:
There are currently two different implementations of header components. We should do the following to consolidate these header components:

1. Replace Header child components (e.g., `HeaderMenu`, `HeaderNavigation`, `HeaderSecondaryNavigation`) with `BaseHeader` child components where applicable. So, for example, `BaseHeaderMenu` would be renamed `HeaderMenu` and the existing `HeaderMenu` would be removed.
2. Make `Header` and its related components use the `BaseHeader` components and provide ordering and processing of those components in this order when any of the following props are present: navigation, search, and avatar.
3. Document that `BaseHeader` and related components are unstructured, unprocessed versions of the Header components if you want more flexibility in ordering components.

johglove:
Completed with merge of https://github.com/indiana-university/rivet-react/pull/486

PR #486, opened 2025-01-09
feat: Merged and cleaned up header components (#451)

johglove:


CONTRIBUTING.md
```diff
@@ -6,6 +6,8 @@ To get started, fork the [rivet-react repo](https://github.com/indiana-universit
 
 To compile the components:
 
+- Requires Node v22.12.0
```
burnumd, line 9:
```suggestion
- Requires Node v22.12.0 (The "Jod" LTS)
```

src/components/Header/HeaderMenu.jsx
```diff
@@ -7,188 +7,41 @@ import PropTypes from "prop-types";
...
+      {React.Children.map(children, renderChild)}
+    </BaseHeaderMenu>
   );
 };
 
 HeaderMenu.displayName = "Header.Menu";
 HeaderMenu.propTypes = {
-  /** The label of the menu */
-  label: PropTypes.string.isRequired,
+  /** Indicates item is current page if link with url */
```
burnumd, line 35: I don't understand this documentation comment. Can you clarify what it means?

src/components/Header/childUtils.jsx
```diff
@@ -53,10 +57,5 @@ export const renderHeaderNavUnorderedList = (child) => {
     child.props.children,
     renderHeaderNavListItem,
   );
-
-  return (
-    <ul key={uuidv4()} className={"rvt-header-menu__list"}>
-      {listItems}
-    </ul>
-  );
+  return listItems;
```
burnumd, line 60: If we're not doing any processing on these, what is this function doing now?

johglove: This removal isnt removing the processing of the list items; it is just removing the ul container that was being wrapped around the items.

burnumd:
Any idea why the code coverage decreased here? This change should reduce the need for some tests in `Header` but that shouldn't affect coverage.

johglove:
Looks like coverage loss was because removed code contained only usages of event util functions, so functions lost test coverage when header components were refactored.  Cleaned out unused functions

burnumd:
I'm seeing a couple errors on the style guide on this branch: `Warning: Failed prop type: children must only contain li,({` and `Warning: Failed prop type: Cannot read properties of undefined (reading 'type')`.

I couldn't figure out exactly whether they were coming from the MD examples or the code itself, but they should be cleaned up before merging.

burnumd (approved):

Issue #401, opened 2023-09-19
Dialog: Missing role and focus handling


basham:
I'm testing the Dialog component, and I noticed some accessibility issues.

1. `role="dialog"` should always be applied to the dialog. It is missing when [`disablePageInteraction` is enabled](https://github.com/indiana-university/rivet-react/blob/v2.0.0-alpha.2/src/components/Dialog/Dialog.jsx#L142).
2. Focus should move to the dialog element when the dialog opens. Right now, focus remains on the trigger. Focus at least does return to the trigger upon close.
3. Focus should be trapped in the dialog. That means, tabbing from the last interactive item in the dialog moves focus to the first interactive item in the dialog and visa versa.

pcberg:
@basham, these are rivet-react Dialog component bugs. Also, it looks like the focus issues are not present on adobe react-spectrum, so they appear to be an issue with our Dialog component (FWIW: Admittedly I didn’t test how the focus works)

basham:
For the last couple weeks, I've been exploring if we could use the native `dialog` element, rather than the `react-aria` library to drive the Dialog component behavior. I found it to work fairly well, with minimal issues when testing with macOS Safari + VoiceOver and Windows Firefox + NVDA.

- [Proof of concept (CodePen)](https://codepen.io/basham/pen/WNLWypX)
- [rvt-dialog (CodePen)](https://codepen.io/basham/pen/zYyXyGz)

However, it seems like the current recommendation of the Digital Accessibility Office is to not use this `dialog` element approach for now.

> It is getting closer.
> 
> The major browsers mostly support it. The gotcha is how well the range of assistive tech supports it, too.
> 
> The process by which pretend becomes real in HTML-land is roughly as follows:
> 
> The new element or attribute is proposed and becomes a real thing in whatwg (aka the living HTML standard). Then some of the major browsers might implement their own vision of $thing. Then there will be months to years of PRs to fix defects in those implementations of $thing. Meanwhile (often after a very long pause) the makers of the various assistive technologies may begin to change their products to provide support for the shifting $thing, many of which are going to rely on the browser implementations becoming more stable. Rinse & repeat for a while — months or sometimes years. Then there is usually an announcement in accessibility circles that $thing is now a genuine and viable thing, and can be used well under certain cases. Then we verify that $thing actually works as expected against the list of tech we evaluate using, and make a go/no go decision.
> 
> We evaluate using:  
> 
[...]

PR #487, opened 2025-01-09
fix: #401 Adjust role fix and add focus handling

pcberg:
I adjusted #480, I believe the role needs to be an attribute instead of as part of style.

For the focus work I settled on adding a new library. Adobe react-aria supports focus trapping (and it works) but I couldn't get it to work in our code base. It probably has to do with the ref object we pass into the library. I tried to replicate the adobe react-aria code but it turned out messy and I called it quits before I got it working. Finally I settled on using focus-trap-react which looks rather clean. Of course it's "one more library"...

Long term I hope we can migrate to HTML5 dialog but as we discussed in the last few days it is not accessibility ready.

Let me know what you think, thanks!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@react-aria/overlays": "3.24.0",
         "@react-aria/utils": "3.26.0",
         "classnames": "2.5.1",
+        "focus-trap-react": "11.0.2",
         "rivet-core": "2.8.1",
         "rivet-icons": "3.0.0"
       },
@@ -4874,7 +4875,6 @@
       "version": "19.0.2",
       "resolved": "https://registry.npmjs.org/@types/react/-/react-19.0.2.tgz",
       "integrity": "sha512-USU8ZI/xyKJwFTpjSVIrSeHBVAGagkHQKPNbxeWwql/vDmnTIBgx+TJnhFnj1NXgz8XfprU0egV2dROLGpsBEg==",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "csstype": "^3.0.2"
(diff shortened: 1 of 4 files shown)


johglove (approved):

burnumd (approved):

basham (changes requested):
The code looks fine. However, the following "React Documentation Dialog Examples" are not working:
- Modal dialog
- Confirmation dialog
- Dialog with form

If fixing those are out of scope for this work, then I'll approve the PR.

pcberg:
> The code looks fine. However, the following "React Documentation Dialog Examples" are not working:
> 
>     * Modal dialog
> 
>     * Confirmation dialog
> 
>     * Dialog with form
> 
> 
> If fixing those are out of scope for this work, then I'll approve the PR.
> 
> The other dialog examples are working well with macOS VoiceOver and Safari.

I noticed that too when I started on this ticket. @burnumd , any new insight from your perspective per our conversation earlier this week? Having said that, I'm of the thought to create a new ticket. I'm happy to take a look too, although next week.

burnumd:
I was able to determine the cause of those dialogs not opening. It's the `closeOnOutsideClickOrEscape` property that's being triggered on the button click that opens the dialogs. I think I'd rather file that as a separate bug since it's pre-existing and that way I can track my work on it.

basham (approved):

pcberg:
> I was able to determine the cause of those dialogs not opening. It's the `closeOnOutsideClickOrEscape` property that's being triggered on the button click that opens the dialogs. I think I'd rather file that as a separate bug since it's pre-existing and that way I can track my work on it.

I created https://github.com/indiana-university/rivet-react/issues/491 and merged this one. Thank you!

PR #490, opened 2025-01-10
docs: update contribution docs for new components

burnumd:
Update the documentation to include all the steps necessary to create a new component in the project.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,7 +53,7 @@ You will also need to add an export to the top level component index in `src/com
 export * from "./path/to/component/index.jsx";
 ```
 
-To add a new section to the documentation, edit `styleguide.config.cjs` and add a new `sections` entry. They will typically follow the this pattern.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -30,15 +30,49 @@ npm run start
...
+}
+```
+
+You will also need to add an export to the top level component index in `src/components/index.js`.
+
+```js
+export * from "./path/to/component/index.jsx";
+```
+
+To add a new section to the documentation, edit `styleguide.config.cjs` and add a new `sections` entry. They will typically follow the this pattern.
```
basham, line 56: Replace "the this" with "this".

basham (changes requested):

CONTRIBUTING.md
burnumd: Thanks for catching that!

basham (approved):

Issue #491, opened 2025-01-13
Some "React Documentation Dialog Examples" are not working


pcberg:
Clicking on 

* Modal dialog
* Confirmation dialog
* Dialog with form

when running the project locally appears to be causing the components to close right after opening. We suspect the `closeOnOutsideClickOrEscape` property is being triggered on the button click that opens the dialogs. It does not seem to break any jest tests and doesn't seem to be reproducible in the deployed documentation.

PR #493, opened 2025-01-13
fix: #491 fix dialog examples

burnumd:
In development environments, the modals which allow closing on outside interactions were being immediately closed because the button event which opened them was also triggering a close.

### DIFF
--- a/src/components/Dialog/Dialog.md
+++ b/src/components/Dialog/Dialog.md
@@ -48,7 +48,10 @@ const DialogExample = () => {
 
   return (
     <>
-      <Button onClick={() => setDialogOpen("default-dialog")}>Default dialog</Button>
+      <Button onClick={event => {
+        setDialogOpen("default-dialog");
+        event.stopPropagation();
+      }}>Default dialog</Button>
       <Dialog
         isOpen={dialogOpen === "default-dialog"}
         onDismiss={handleDismiss}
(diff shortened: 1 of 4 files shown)


src/components/Dialog/Dialog.md
```diff
@@ -48,7 +48,10 @@ const DialogExample = () => {
 
   return (
     <>
-      <Button onClick={() => setDialogOpen("default-dialog")}>Default dialog</Button>
+      <Button onClick={event => {
+        setDialogOpen("default-dialog");
+        event.stopPropagation();
```
johglove, line 53: Does `event.preventDefault();` need to be added as well for this just in case? Don't know if there might be a case where unwanted default functionality might trigger but preventDefault/stopPropagation are usually paired together to insure only the provided onClick functionality triggers.

burnumd: I've added that, I also cleaned up the onClick function so it's declared once for readability.

pcberg (approved):
Why did this still work on the deployed documentation (https://indiana-university.github.io/rivet-react/beta/#/Overlays?id=dialog), just not our laptops?

johglove (approved):

burnumd:
@pcberg I couldn't find any open issues or documentation on the styleguidist repository, but it seems to behave differently when you use `styleguidist server` versus `styleguidist build`. One known difference is that it uses Webpack in development mode rather than doing a production build of the library, which likely causes the behavioral differences.

Issue #497, opened 2025-01-14
Need Support for Full Width Buttons


johglove:
Current Button only supports a size of default (no size attribute provided) or 'small'.  Attempting to put anything else results in 

Ve.setExtraStackFrame is not a function
TypeError: Ve.setExtraStackFrame is not a function
    at Se (http://localhost:8080/static/js/bundle.js:172024:12)
    at Mr (http://localhost:8080/static/js/bundle.js:172041:448)
    at Xr (http://localhost:8080/static/js/bundle.js:172245:11)
    at tr (http://localhost:8080/static/js/bundle.js:172305:34)
    at Object.et [as jsx] (http://localhost:8080/static/js/bundle.js:172312:14)

johglove:
Corrected by https://github.com/indiana-university/rivet-react/pull/498

PR #498, opened 2025-01-14
feat: Added support for full width buttons (#497)

johglove:


src/components/Button/Button.jsx
```diff
@@ -34,7 +34,18 @@ const buttonStyle = (variant, loading, modifier) => {
...
  * @see https://rivet.uits.iu.edu/components/forms/buttons/#small-buttons
  */
-const buttonSize = (size) => (size ? `${buttonClass}--${size}` : undefined);
+const buttonSize = (size) => {
+  if (!size) {
+    return undefined;
+  }
+
+  switch (size) {
+    case "full":
```
burnumd, line 43: If `small` and `full-width` are mutually exclusive why not use `full-width` as the property name so there's less business logic in here?

burnumd (approved):

PR #506, opened 2025-01-21
fix: inline rivet icons

burnumd:
Until [rivet-icons/#109](https://github.com/indiana-university/rivet-icons/issues/109) is complete, include a modified version of the `rivet-icon-element.js` which registers the icons as `rvt-react-icon` tags. Client applications can then use rivet-icons as documented.


johglove (approved):

basham:
Overall, this seems like a reasonable short-term solution.

basham (approved):

Issue #510, opened 2025-01-23
Allow Setting of the Element for Grid Components


johglove:
Currently the element for all grid components is div.  While this would work for most cases there are situations when developers may wish to use other elements such as article, section, etc.  Add the ability for users to set the element of grid components.

johglove:
Released on https://github.com/indiana-university/rivet-react/releases/tag/v2.1.0

PR #511, opened 2025-01-23
feat: Allow setting of Grid component elements (#510)

johglove:


src/components/Grid/Col.md
```diff
@@ -123,3 +123,23 @@ The `last` property can pull a column to the right
 </Container>
 ```
 <!-- prettier-ignore-end -->
+
+### Set elements of containers
+
```
burnumd, line 128: It would be worthwhile to have additional documentation about why users would want to do this and what it means to set the `component` prop.

burnumd (changes requested):
Can you edit the title and description of this PR to reflect what's changing?

src/components/Grid/Col.md
```diff
@@ -123,3 +123,25 @@ The `last` property can pull a column to the right
...
+The `component` property overrides the element of the container allowing for the user of semantic elements when needed
```
burnumd, line 129:
```suggestion
The `component` property overrides the element of the container allowing for the use of semantic elements when needed. The default component is a `div`.
```

burnumd (changes requested):

src/components/Grid/Col.jsx
```diff
@@ -73,6 +77,8 @@ Col.propTypes = {
   /** Width of the column (1-12) */
   // prettier-ignore
   columnWidth: PropTypes.oneOf(validWidths),
+  /** Sets the containing element. Defaults is 'div' */
```
burnumd, line 80:
```suggestion
  /** Sets the containing element. The default container element is a 'div' */
```

burnumd (changes requested):

basham (approved):

burnumd (approved):

Issue #512, opened 2025-01-24
Unable to have an Ordered and Plain List Using the List Component


johglove:
The List component variant property allows for 3 settings 'ordered', 'unordered' or 'plain' with 'unordered' being the default and only 1 value accepted.  However this controls 2 different changes in the component.  Unordered and Ordered dictate the element of the container 'ul' vs 'ol' with 'ul' as default.  Plain simply adds an additional className value of 'rvt-list-plain'.  Possible combinations:

1. ul with no class added (variant='unordered')
2. ul with rvt-list-plain added (variant='plain')
3. ol with no class added (variant='ordered')
4. ol with rvt-list-plain added (Not Possible)

Current work around would be ordered variant with className manually set but is not user friendly and with description list already split out the value of using the component is already minimal.

basham:
This issue should also include consideration of the `rvt-list-inline` class.

What if these are the values (and the result) for the `variant` prop?

- `description` (`dl.rvt-list-description`)
- `ordered` (`ol.rvt-list`)
- `ordered-inline (`ol.rvt-list-inline`) 
- `ordered-plain (`ol.rvt-list-plain`)
- `unordered` (`ul.rvt-list`)
- `unordered-inline (`ul.rvt-list-inline`)
- `unordered-plain` (`ul.rvt-list-plain`)

Also, I see a `rvt-list-reset` class. I've never used it. It is not documented, so maybe we shouldn't support it. But if we would add it, maybe we'd need these, too?

- `ordered-reset` (`ol.rvt-list-reset`)
- `unordered-reset` (`ul.rvt-list-reset`)

I could see later contributing a new [key/value variant](https://codepen.io/basham/pen/rNyQXjW) for description lists (whatever it would be called). So, that could be added like:

- `description-key` (`dl.rvt-list-description-key`)

burnumd:
There's a workaround for this issue and the fix for it would be a breaking change, so this issue is being closed as won't fix but the solution will be implemented in 3.x.

PR #513, opened 2025-01-24
fix: Refactored List component so type and display options are handled separately (#512)

johglove:


src/components/List/List.test.jsx
```diff
@@ -41,34 +41,96 @@ describe("<List />", () => {
...
           <li>Three</li>
-        </List>
+        </List>,
       );
       const list = await screen.findByTestId(testId);
       expect(list.nodeName).toBe("OL");
     });
 
-    it("should render a plain list if variant is plain", async () => {
+    it("should render a description list if variant is ordered", async () => {
```
burnumd, line 50:
```suggestion
    it("should render a description list if variant is description", async () => {
```

burnumd (changes requested):

src/components/List/List.jsx
```diff
@@ -18,15 +18,17 @@ const List = ({
...
   children,
+  plain = false,
   variant = "unordered",
   ...attrs
 }) => {
   const classes = classNames(
     inline ? `${classPrefix}-inline` : classPrefix,
-    variant === "plain" && `${classPrefix}-plain`,
-    className
+    plain ? `${classPrefix}-plain` : "",
```
burnumd, line 27:
```suggestion
    (plain || variant === "plain") ? `${classPrefix}-plain` : "",
```

Would help preserve backwards compatibility if the variant plain is deprecated and not actually removed.

burnumd (commented):
I'm not sure this change qualifies as a `fix` since it will break existing uses of the `List` component. I made a suggestion to preserve the functionality, but I'm curious of others' opinions.

src/components/List/List.jsx
```diff
@@ -18,15 +18,18 @@ const List = ({
...
   const classes = classNames(
     inline ? `${classPrefix}-inline` : classPrefix,
+    plain ? `${classPrefix}-plain` : "",
     variant === "plain" && `${classPrefix}-plain`,
-    className
+    variant === "description" && `${classPrefix}-description`,
+    className,
   );
-  const ListTag = variant === "ordered" ? "ol" : "ul";
+  const ListTag = getElement(variant);
```
basham, line 32: Rather than a function, you could do this:

```js
// At the top of the module:
const variantTagMap = {
  description: 'dl',
  ordered: 'ol',
  plain: 'ul',
  unordered: 'ul'
};

// Here:
const ListTag = variantTagMap[variant];
```

Up to you if you want to change it. But I feel like these mapping objects tend to be cleaner than functions and switch statements.

basham: Also, I wonder if we need to be a little more internally consistent on something. Here, we use both "tag" and "element" interchangeably. Then PR #511 uses the "component" prop. Maybe that's something we clean up at some point.

basham: Likewise, HTML conventions tends to use a "type" attribute similarly to how we're using the "variant" prop here. So again, I don't know if this is something to discuss.

src/components/List/List.jsx
basham: We shouldn't have a "plain" prop, because then it could lead to some weird situations:

```jsx
<List plain variant="description">
```

There should just be the "variant" prop.

basham (changes requested):
Don't consider my comments a blocker.

src/components/List/List.jsx
burnumd: I agree with standardizing on `component` (I've seen that on other projects). HTML using `type` as an attribute throws a wrench in our library components.l I think it was alerts we had a problem with in the alpha and switched away from `type`.

johglove:
Having plain as a variant is what caused the original issue (https://github.com/indiana-university/rivet-react/issues/512).  While having it as a separate prop may cause strange combinations they don't directly conflict outside and it would be on the users if they want the combination and if so how to handle any strangeness.

src/components/List/List.jsx
johglove: Having plain as a variant is what caused the original issue (https://github.com/indiana-university/rivet-react/issues/512). While having it as a separate prop may cause strange combinations they don't directly conflict outside and it would be on the users if they want the combination and if so how to handle any strangeness.

johglove:
My original intent was to keep existing functionality but just in a deprecated state until a major release. However forgot I needed to keep the original line and have readded it.

src/components/List/List.jsx
basham: FYI: I added a comment to #512 that I'm hoping may inform what happens in this PR.

Issue #514, opened 2025-01-28
Unable to Set Id Property to Dialog Components


johglove:
The Dialog component takes an id property but it is only used to set the title id which is then used for the aria label by attribute.  It does not appear on the dialog container or other components. though it looks like it should be set for it.  This may because of overwriting done by the the library's useDialog/useModalOverlay functions as the setting of the header titles has removes the id from the processed props.

PR #515, opened 2025-01-28
fix: Set ids to Dialog container and close button (#514)

johglove:


src/components/Dialog/Dialog.md
```diff
@@ -55,6 +56,7 @@ const DialogExample = () => {
     <>
       <Button onClick={event => openModal(event, "default-dialog")}>Default dialog</Button>
       <Dialog
+        id='Default-Dialog'
```
burnumd, line 59:
```suggestion
        id="Default-Dialog"
```

src/components/Dialog/Dialog.md
```diff
@@ -16,6 +16,7 @@ const DialogExample = () => {
     <>
       <Button onClick={() => setDialogOpen(true)}>Open Dialog</Button>
       <Dialog
+        id='Example-Dialog'
```
burnumd, line 19:
```suggestion
        id="Example-Dialog"
```

src/components/Dialog/Dialog.jsx
```diff
@@ -142,7 +143,10 @@ const Dialog = ({
         )}
         {children}
         {onDismiss && showCloseButton && (
-          <DialogCloseButton onDismiss={handleDismiss} />
+          <DialogCloseButton
+            id={`${id}-close__button`}
```
burnumd, line 147: Since ID is not marked as required, we could run into issues with this construction (namely, having a bunch of IDs called `undefined-close__button`.

burnumd (changes requested):

src/components/Dialog/Dialog.jsx
johglove: If the id isn't supplied the default is to generate one using `Rivet.shortuid()`

PR #521, opened 2025-02-10
feat: Added callback support when changing tabs (#520)

johglove:


burnumd (commented):
This all looks fine to me, but I think it should be a feature rather than a fix.

johglove:
As the lack of the functionality doesn't lead to a break it this would be more an enhancement than a bug.  However feature releases are marked as major changes (ie 2.1.x -> 2.2.x) and as is this enhancement doesnt feel like that level of change.  As a patch change to fix missing functionality fits better (2.1.2 -> 2.1.3)

burnumd:
Looking at the [SemVer documentation](https://semver.org/), minor versions are intended for

> MINOR version when you add functionality in a backward compatible manner

Which seems to be what this change is.

PR #523, opened 2025-02-11
feat: Added customizable rivetize component (#522)

johglove:


burnumd (approved):

src/components/util/RivetizeComponent.jsx
```diff
@@ -0,0 +1,26 @@
+/*
+Copyright (C) 2018 The Trustees of Indiana University
+SPDX-License-Identifier: BSD-3-Clause
+*/
+import classNames from "classnames";
```
basham, line 5: This dependency isn't doing anything. Should it? If not, remove it.

src/components/util/RivetizeComponent.jsx
```diff
@@ -0,0 +1,26 @@
...
+import * as React from "react";
+import * as PropTypes from "prop-types";
+
+import * as Rivet from "./Rivet";
```
basham, line 9: This dependency isn't doing anything. Should it? If not, remove it.

basham (changes requested):

Issue #525, opened 2025-02-19
Update rivet-icons and rivet-stickers


burnumd:
Both [rivet-icons](https://github.com/indiana-university/rivet-icons/releases/tag/v3.0.1) and [rivet-stickers](https://github.com/indiana-university/rivet-stickers/releases/tag/v0.3.1) have had updated releases which fix an issue we were having defining icons/stickers in client applications. We should bring in these new releases and update rivet-react to handle the new paradigm.

PR #526, opened 2025-02-19
feat: #525 update rivet icons and stickers

burnumd:
Upgrade dependencies including rivet-icons and rivet-stickers. Removing the rvt-react-icon component allows us to use the upgraded javascript behind rivet-icons/stickers and client applications can rely on these elements only being registered once.
Note that this change includes making the icons and stickers packages peer dependencies. This requirement was already in place because clients need to `import` the CSS and Js files from those packages, this change just formalizes that requirement.


README.md
```diff
@@ -39,6 +39,17 @@ import { Alert } from "rivet-react";
 <Alert variant="info">A very important message for you!</Alert>;
 ```
 
+## Using [rivet-icons](https://github.com/indiana-university/rivet-icons) and [rivet-stickers](https://github.com/indiana-university/rivet-stickers)
+
+Rivet-react relies on several icons and stickers from the Rivet Icons and Rivet Stickers packages. In order to load the CSS and custom JavaScript necessary to display these elements, client applications will need to include the following in their `index.jsx` file similarly to how the Rivet core CSS is imported.
+
+```typescript static
+import "~/rivet-icons/dist/rivet-icon-element.js";
```
basham, line 47: Technically, `rivet-icon-element.js` and `rivet-sticker-element.js` don't have to be included in the client side for v3.0.1 and v0.3.1, respectively. But since the plan is to make it a requirement in the next major version of those packages — I'm fine with how this is worded.

basham (approved):

johglove (approved):

Issue #527, opened 2025-02-27
Right Align Last Column Missing from Col Component


johglove:
https://rivet.uits.iu.edu/components/grid/?example=right-align-last-column

PR #528, opened 2025-02-27
fix: Added missing prop type for Col component (#527)

johglove:


burnumd (approved):

src/components/Grid/Col.jsx
```diff
@@ -79,6 +79,8 @@ Col.propTypes = {
   columnWidth: PropTypes.oneOf(validWidths),
   /** Sets the containing element. The default container element is a 'div' */
   component: PropTypes.string,
+  /** If last should the column be right aligned.*/
```
basham, line 82: This phrasing is odd. It reads as if it is a question. And is it about right text alignment or putting it at the end of the row/block?

basham (changes requested):

src/components/Grid/Col.jsx
basham: Maybe: "Place the column at the end of the row."

basham (approved):

PR #530, opened 2025-02-27
Feature/529

johglove:


src/components/Grid/Col.jsx
```diff
@@ -6,27 +6,55 @@ import classNames from "classnames";
 import * as PropTypes from "prop-types";
 import * as React from "react";
 import * as Rivet from "../util/Rivet";
+import { Edges, Sizes } from "../util/RivetizeConstants.js";
```
basham, line 9: These imports are not used (as far as I can tell). Remove them.

basham (approved):

PR #532, opened 2025-02-28
fix: Added handling for numbers for Col columnWidth (#531)

johglove:


src/components/Grid/Col.jsx
```diff
@@ -29,7 +29,7 @@ const getColumnClass = (columnWidth, breakpoint) => {
   if (!(columnWidth || breakpoint)) {
     return columnClassPrefix;
   }
-  if (typeof columnWidth === "string") {
+  if (typeof columnWidth === "string" || typeof columnWidth === "number") {
```
basham, line 32: Do the prop types need to change at all?

basham (approved):

src/components/Grid/Col.jsx
johglove: The columnWidth previously allowed string and numbers before the change to allow multiple values set.  The string and number values were already listed in the valid values for the prop types (object was added by multiple value work).

PR #538, opened 2025-03-17
feat: EPMDW-3684 - Allow node for AccordionPanel title prop - Closes #537

### LINKED ISSUE #537: Allow AccordionPanel title prop to take element
It is currently restricted to string

athill:


burnumd (commented):
I don't have any objection to the content of this PR, but I think it should be a feature rather than a fix since it enables new functionality and the existing behavior worked correctly before.

athill:
> I don't have any objection to the content of this PR, but I think it should be a feature rather than a fix since it enables new functionality and the existing behavior worked correctly before.

Commit type has been updated. Thanks.

burnumd (approved):

PR #554, opened 2025-08-11
docs: fix Header.Menu documentation

burnumd:
The `label` property is not actually required for header menu items. Also, update the documentation to wrap menu items in a `Header.Menu` component in the `Header` documentation so links are clickable.

### DIFF
--- a/src/components/Header/HeaderMenu.jsx
+++ b/src/components/Header/HeaderMenu.jsx
@@ -36,7 +36,7 @@ HeaderMenu.propTypes = {
   /** The href that the menu's label links to, wraps label in anchor if provided*/
   href: PropTypes.string,
   /** The label of the menu */
-  label: PropTypes.string.isRequired,
+  label: PropTypes.string,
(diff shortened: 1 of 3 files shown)


src/components/Header/HeaderMenu.jsx
```diff
@@ -36,7 +36,7 @@ HeaderMenu.propTypes = {
   /** The href that the menu's label links to, wraps label in anchor if provided*/
   href: PropTypes.string,
   /** The label of the menu */
-  label: PropTypes.string.isRequired,
+  label: PropTypes.string,
```
athill, line 39: Do we also want to make this a `node`? Might save us from needing to do it later.

athill (approved):
Looks good overall. One potential change.

src/components/Header/HeaderMenu.jsx
athill: I see. Makes sense.

PR #555, opened 2025-08-14
fix: Corrected trigger for mobile/collapsed nav menu

johglove:
The issue was the check for should the menu toggle would detect the click as outside and close the menu immediately after open if the icon element in the button was clicked.   This is because the element contains method does not work with the icons due to render timing.  For this modified the setup of the menu to match the way the Header Menu works as it fixed the issue and added consistency in the menus.

### DIFF
--- a/src/components/Header/BaseHeaderNavigation.test.jsx
+++ b/src/components/Header/BaseHeaderNavigation.test.jsx
@@ -107,28 +107,6 @@ describe("<BaseHeaderNavigation />", () => {
       ).toHaveAttribute("name", "close");
     });
 
-    it("should not hide the nav if the Tab key is pressed while the nav is open", async () => {
-      expect(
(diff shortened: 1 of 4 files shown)


src/components/Header/BaseHeaderNavigation.test.jsx
```diff
@@ -107,28 +107,6 @@ describe("<BaseHeaderNavigation />", () => {
       ).toHaveAttribute("name", "close");
     });
 
-    it("should not hide the nav if the Tab key is pressed while the nav is open", async () => {
```
johglove, line 110: Because tabbing outside the menu (from the last item or shift+tab from the first) moves focus outside of the menu it should cause the menu to close to prevent loss/cover of focus, making this test unneeded.

PR #557, opened 2025-08-25
chore: Update Dependencies. Support React 19, latest Vite. fix #556

### LINKED ISSUE #556: Dependency updates: Support React 19
Update dependencies. Support latest Vite and React 19.

james-anderson-iu:
- Updates dependencies.
- It's not a guarantee of backwards compatibility, but I tested it in a template based on a basic Vite project using React 18.3 and React 19. It worked with both.
- I ran build, test-win, pretty-quick, prepare, license, and start successfully.
- There were some changes that don't seem like they should be necessary, but caused errors during tests or when testing it with the template Vite project. If someone else tries these updates without those file changes and is able to run all the npm scripts without error and test with a React 18.3 project and a React 19 project, maybe those changes can be removed.
- Can we include sourcemaps? If not, that change can be removed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,11 +59,15 @@
     },
     "node_modules/@adobe/css-tools": {
       "version": "4.4.4",
+      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.4.4.tgz",
+      "integrity": "sha512-Elp+iwUx5rN5+Y8xLt5/GRoG20WGoDCQ/1Fb+1LiGtvwbDavuSk0jhD/eZdckHAuzcDzccnkv+rEjyWfRx18gg==",
       "dev": true,
       "license": "MIT"
     },
     "node_modules/@ampproject/remapping": {
       "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
+      "integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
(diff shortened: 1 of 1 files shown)


burnumd (changes requested):
Thanks for the pull request. These changes do seem to break backwards compatibility with React 18 projects on my machine. I'm going to take a closer look and try to find what's happening.

james-anderson-iu:
I added more rivet-react components in the project I was testing with and re-did the `npm install` from scratch. The main thing I needed to update on the test project was to add dependencies for all of rivet-react's dependencies that are in `rollupOptions.external` in the vite config. I did update all the test project's dependencies except React and ReactDOM too. Here's what the dependencies look like in my test project. I also tried with all the dev dependencies of the test project reverted back to their old versions and a fresh npm install.
```
"dependencies": {
    "@react-aria/dialog": "3.5.29",
    "@react-aria/overlays": "3.29.0",
    "@react-aria/utils": "3.30.1",
    "classnames": "2.5.1",
    "focus-trap-react": "11.0.4",
    "react": "18.3.1",
    "react-dom": "18.3.1",
    "rivet-core": "2.9.1",
    "rivet-icons": "3.0.1",
    "rivet-react": "file:../rivet-react/rivet-react-0.0.0.tgz",
    "rivet-stickers": "0.3.1",
    "uuid": "11.1.0"
  },
```
There was one odd thing. I had a cypress test where I had to add `{force: true}` to a step that checked a checkbox. I didn't see any reason why cypress thought the element wasn't able to be interacted with at that point.

When I was initially testing, I used `"rivet-react": "file:../rivet-react"` as a dependency in the test project to test changes made to rivet-react with the test project. Once it seemed like the rivet-react changes were going to work, I tried running `npm pack` and using the rivet-react archive as a dependency, to mimic installing it from something equivalent to an published version on npm.

PR #560, opened 2025-10-06
feat: update build tools to support React 19

burnumd:
Update Vite (and associated plugins) and the Vite rollup configuration to support React 19 as a peer dependency.


athill (approved):

burnumd:
@james-anderson-iu I think I'm going to file an issue on the project to convert to TypeScript and close this PR. @johglove  @basham @pcberg, I'll schedule a quick meeting to discuss if/how we want to proceed with that change. We should be able to do it incrementally.

PR #591, opened 2026-04-28
fix: Updated Dependencies to Allow Use of React 19 (#582)

johglove:


src/components/PageContent/Billboard/Billboard.test.jsx
```diff
@@ -182,5 +177,7 @@ const checkRenderImage = () => {
   const billboardImage = screen.getByTestId(testIds.image);
   expect(billboardImage).toBeVisible();
   expect(billboardImage).toHaveClass("rvt-billboard__image");
-  expect(billboardImage.innerHTML).toBe(imageStr);
+  expect(billboardImage.innerHTML).toContain("img");
+  expect(billboardImage.innerHTML).toContain(`src="${src}"`);
```
athill: I believe you could do something like:
```
const image = within(billboardImage).querySelector('img');
expect(image.src).toBe(src);
expect(image.alt).toBe(alt)
```

athill (approved):
Looks good.

burnumd (approved):

PR #620, opened 2026-06-16
fix: Updated Dependencies to Allow Use of React 19 (#582)

johglove:


src/components/Header/BaseHeaderMenu.jsx
```diff
@@ -120,7 +120,11 @@ const BaseHeaderMenu = ({
         )}
         {!menuUrl && (
           <span
-            data-testid={testMode ? TestUtils.Header.menuAnchorTestId : null}
+            {...(testMode
+              ? {
+                  "data-testid": TestUtils.Header.menuAnchorTestId,
+                }
+              : {})}
```
burnumd, line 127: This looks like a reversion from the merge commit. The ternary approach is recommended over spreading a potentially false boolean, which can cause errors.

burnumd (changes requested):
I didn't review `package-lock.json`, but you installed after doing the upstream merge into your branch, right?

src/components/Header/BaseHeaderMenu.jsx
johglove: How would a boolean be used? 
If testMode is true then {"data-testid": TestUtils.Header.menuAnchorTestId,} is used, if false then {}.
Spread operation is used to reduce markups by preventing test attributes in elements.

burnumd: Hm, it may be an artifact of how Jest was operating in the GHA workflows when I did the GHA conversion, but I ran into test failures because of it.

johglove: Was it possibly the && method that was giving the errors.  I had to change to providing a false case with inline conditional to prevent errors (don't know what started them as it was with unchanged codebase).

burnumd (approved):

johglove:
I cleared both package lock and my node modules and did a fresh install after the merging from master to create a fresh lock file.

PR #631, opened 2026-07-30
fix: apply aria attributes to all dialogs

### LINKED ISSUE #630: Dialog component does not accurately populate aria attributes
The Dialog component currently conditionally provides either `dialogProps` or `modalProps` based on whether the background can be interacted with. That, combined with the interactions with react-aria, cause the appropriate aria attributes to be assigned to the dialog header.

burnumd:
Require a title for all dialog components (since they are required for accessibility) and properly populate `aria-labelledby` attributes for the surrounding div.

Additionally another test was throwing a warning that `div` cannot accept Javascript boolean attributes the way a React component can.

Closes #630

### DIFF
--- a/src/components/Dialog/Dialog.jsx
+++ b/src/components/Dialog/Dialog.jsx
@@ -33,7 +33,7 @@ const Dialog = ({
   isOpen,
   onDismiss,
   showCloseButton = true,
-  title,
+  title = "Dialog",
   ...attrs
 }) => {
   const ref = React.useRef(null);
@@ -137,7 +137,7 @@ const Dialog = ({
             // don't want the id from react-spectrum Dialog, use rivet format
             {...removeProperty(titleProps, "id")}
           >
-            {title || "Dialog"}
+            {title}
           </h1>
         </header>
         {children}
--- a/src/components/Dialog/Dialog.jsx
+++ b/src/components/Dialog/Dialog.jsx
@@ -119,6 +119,7 @@ const Dialog = ({
     >
       <div
         className={classNames("rvt-dialog", className)}
+        aria-labelledby={`${id}-title`}
         aria-hidden={!isOpen}
         hidden={!isOpen}
         ref={ref}
@@ -126,21 +127,19 @@ const Dialog = ({
         {...overlayProps}
         id={id}
       >
-        {title && (
-          <header
-            className="rvt-dialog__header"
-            data-testid={TestUtils.Dialog.dialogHeaderTestId}
+        <header
+          className="rvt-dialog__header"
+          data-testid={TestUtils.Dialog.dialogHeaderTestId}
+        >
+          <h1
+            className="rvt-dialog__title"
+            id={`${id}-title`}
+            // don't want the id from react-spectrum Dialog, use rivet format
+            {...removeProperty(titleProps, "id")}
           >
-            <h1
-              className="rvt-dialog__title"
-              id={`${id}-title`}
-              // don't want the id from react-spectrum Dialog, use rivet format
-              {...removeProperty(titleProps, "id")}
-            >
-              {title}
-            </h1>
-          </header>
-        )}
+            {title || "Dialog"}
+          </h1>
+        </header>
         {children}
         {onDismiss && showCloseButton && (
           <DialogCloseButton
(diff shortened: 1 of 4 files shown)


src/components/Dialog/Dialog.jsx
```diff
@@ -119,28 +119,27 @@ const Dialog = ({
...
-              className="rvt-dialog__title"
-              id={`${id}-title`}
-              // don't want the id from react-spectrum Dialog, use rivet format
-              {...removeProperty(titleProps, "id")}
-            >
-              {title}
-            </h1>
-          </header>
-        )}
+            {title || "Dialog"}
```
johglove: If adding a default then title isn't exactly required.

src/components/Dialog/Dialog.jsx
```diff
@@ -119,28 +119,27 @@ const Dialog = ({
     >
       <div
         className={classNames("rvt-dialog", className)}
+        aria-labelledby={`${id}-title`}
```
johglove, line 122: Does the aria-labelledby need to be on the same element as role='dialog'?  If so the version using useModalOverlay has role='dialog' on the wrapping container and this would need to be moved to it.  The version built with useDialog would have role on this element with the aria-labelledby here being overwritten by the passed in overlayProps.

burnumd: That was confusing to me. I think it stems from using `react-aria`. The Rivet docs have the `aria-labelledby` attribute on the `div` with the `rvt-dialog` class and `role="dialog`. The `role="dialog"` div appears to just be used to allow `FocusTrap` to identify the element that should trap focus. I'm going to push another commit that uses a unique id for the focus trap query selector and adds the `role="dialog"` attribute to the `dialogContent` outer `div`.

src/components/Header/BaseHeaderNavigation.test.jsx
```diff
@@ -397,7 +397,7 @@ describe("<BaseHeaderNavigation />", () => {
         <BaseHeaderNavigation testMode>
           <BaseHeaderMenuItem>Nav item one</BaseHeaderMenuItem>
           <li>Nav item two</li>
-          <div navlistitem>Nav item three</div>
+          <div navlistitem="true">Nav item three</div>
```
athill: I'm not sure how this is working, I would think `div` tags could only use [valid HTML attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes). 

OK, I see [what is happening](https://github.com/indiana-university/rivet-react/blob/master/src/components/Header/BaseHeaderNavigation.jsx#L98), but I wonder if 
a) Could we modify the logic in the component to something like `'navlistitem' in child.props && child.props.navlistitem !== false` to allow the attribute without an explicit value?
b) Can we strip the `navlistitem` attribute from the `div` before rendering? It's still not a valid HTML attribute
c) Unless it is the only way to do it, passing the string "true" to `navlistitem` feels wrong. I'm pretty sure `navlistitem="false"` would also register as truthy, as it is a non-empty string

athill (changes requested):
Looks good overall. I appreciate your addressing the warnings we receive. Just a few questions about the change to `BaseHeaderNavigation.test.jsx`

src/components/Dialog/Dialog.jsx
basham: Assistive tech ignores ARIA labels if they are not attached to an element with a role. With few exceptions, every role should have a label (implicit or explicit). For dialogs, their labels should be attached with `aria-labelledby`.

src/components/Header/BaseHeaderNavigation.test.jsx
basham: Custom attributes should be prefixed with `data-*`. This is optional for custom elements (since the dev is the one designing its API).

basham: It may be more clear what's happening if the `<div>` was instead a React component.

athill: I think the idea is that you can use arbitrary elements, which is why I suggested removing the prop/attribute before rendering.

basham: Yes, agreed.

src/components/Dialog/Dialog.jsx
burnumd: Right, so in this case the `role="dialog"` is being set by react-aria, so we need the `aria-labelledby` attribute here.

johglove: Since most tech ignores aria if not on the role element the double up when disablePageInteraction = true shouldn't be an issue (aria-labelledby would be on both this div and the wrapper div with wrapper having the role dialog).
The only case where role is on this element is when disablePageInteraction = false and dialogProps is pass into dialogContent.  From the looks of line 51, dialogProps will have both ["aria-labelledby"]: `${id}-title`, role: "dialog" which will overwrite this line since.

Having it here shouldn't hurt anything but don't think it is needed.

johglove (approved):

athill (approved):
Looks good. Thanks!

PR #635, opened 2026-08-06
fix/625 - Added deprecation labeling

johglove:


src/components/Footer/ResourceFooter.jsx
```diff
@@ -20,6 +20,11 @@ const ResourceFooter = ({
   size = "sm",
   ...attrs
 }) => {
+  React.useEffect(() => {
+    console.warn(
+      "ResourceFooter is deprecated and will be removed in a future release.",
+    );
+  }, []);
```
burnumd: What if this were extracted to a custom hook, `useDeprecation(componentName)` to abstract this functionality so there's a consistent warning message if other components are deprecated in the future?

src/components/Footer/ResourceFooterLinkBlock.md
```diff
@@ -1,4 +1,4 @@
-### Resource Footer Link Block
+(Deprecated) ### Resource Footer Link Block
```
burnumd: Is there a reason the deprecation text is before the header tag? Do these warnings even need to be added directly to the markdown if the component has a jsdoc `@deprecated` flag in the component documentation?

burnumd (changes requested):

